PR #6, opened 2023-05-25
fix PROTOCOL env var

mraszyk:


constants.sh
```diff
@@ -82,7 +82,7 @@ fi
 
 export NETWORK=$([[ "$TESTNET" == "local" ]] && echo "local" || echo "https://${TESTNET}")
 export DFX_NETWORK=$([[ "$TESTNET" == "local" ]] && echo "local" || echo "https___${TESTNET//./_}")
```
aterga, line 84: Why is this `___` okay?

mraszyk: The `___` is the right choice because dfx replaces special symbols with `_` and uses such networks at certain places.

PR #9, opened 2023-05-26
change the directory to repository's root in scripts

mraszyk:
This PR makes the following changes:
- allows all scripts to be executed from anywhere (changing the directory to the root of sns-testing at the beginning of every script; note that this does not affect the bash session executing the script unless it is _sourced_, e.g., by running `source open_sns_sale.sh`)
- make `upgrade_dapp.sh` run `dfx build` commands from where the script is executed, e.g., the project-specific `/dapp` directory, and accept a path to a file containing binary-encoded post-upgrade arguments
- add `developer_neuron_id.sh` script printing the developer neuron ID to stdout
- use `idl2json`, `jq`, and `python3` to derive SNS neuron IDs


developer_neuron_id.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
+
+set -euo pipefail
+
+. ./constants.sh normal
+
+echo "$(dfx canister --network "${NETWORK}" call sns_governance list_neurons "(record {of_principal = opt principal\"${DFX_PRINCIPAL}\"; limit = 1})" | grep "^ *id = blob" | sed "s/^ *id = \(.*\);$/'(\1)'/" | xargs didc encode | tail -c +21)"
```
aterga, line 9: Nit: I would prefer to use `didc` and `jq` to select fields from inherently structured data.

aterga:
Has this been tested (1) with the docker solution, (2) the native- macOS instructions, or both?

developer_neuron_id.sh
mraszyk: Using `idl2json`, `jq`, and `python3` to derive SNS neuron IDs now.

mraszyk:
> Has this been tested (1) with the docker solution, (2) the native- macOS instructions, or both?

I've successfully run the basic scenario in docker at the latest commit 4032ace. I haven't tested on macOS as I don't have one.

aterga:
> > Has this been tested (1) with the docker solution, (2) the native- macOS instructions, or both?
> 
> I've successfully run the basic scenario in docker at the latest commit [4032ace](https://github.com/dfinity/sns-testing/commit/4032ace37e58b46984a8ef64a309c01988ee41a9). I haven't tested on macOS as I don't have one.

Great, thanks. Then I'll test on Apple Silicon, and if we works as expected, I think we can merge this MR.

aterga:
Progress so far:
* If the 8080 port is occupied, the first step inside Nix will fail. One should run `docker ps` _and_ `lsof -i :8080` to determine which process or container needs to be terminated.
* `Basic scenario has successfully finished.` !

aterga:
Nit: step 6 should provide sane example values for `./participate_sns_swap.sh <num-participants> <icp-per-participant>`

aterga:
Step 6 in "SNS lifecycle" 

Running 

```
[nix-shell:~/sns-testing]$ ./participate_sns_swap.sh 2 10
Using identity: "default".
Using identity: "default".
Error: Identity already exists.
Using identity: "participant-000".
Using identity: "icp-ident-RqOPnjj5ERjAEnwlvfKw".
Transfer sent at block height 24
Using identity: "participant-000".
...
```
I was wondering if the error indicates that something didn't go as expected. If it's expected, maybe we could mention that as a not in the README instructions.

aterga:
Also step 6:

How can I get the account of e.g., `participant-001`?

aterga:
Step 7: there should be a note explaining that the sale finalization will expectedly fail if the participation is insufficient: 

```
(
  record {
    set_dapp_controllers_call_result = null;
    settle_community_fund_participation_result = null;
    error_message = opt "The Sale can only be finalized in the COMMITTED or ABORTED states. Current state is Open";
    set_mode_call_result = null;
    sweep_icp_result = null;
    claim_neuron_result = null;
    sweep_sns_result = null;
  },
)
```

aterga:
Is it safe to run the `./participate_sns_swap.sh` multiple times?

aterga (approved):
Basic scenario works on Apple M1. Thanks!

mraszyk:
Thank you for the feedback! It's addressed in https://github.com/dfinity/sns-testing/pull/21 and https://github.com/dfinity/sns-testing/pull/26.

PR #14, opened 2023-06-08
Update README.md

letmejustputthishere:


README.md
```diff
@@ -79,14 +79,14 @@ The `sns-testing` solution is based on Docker; however, there are subtle issues
    ```
    and run the setup script:
    ```bash
-   bash-3.2$ ./setup_locally.sh
```
mraszyk, line 82: The prefix `bash-3.2` is to emphasize that we run the script from within a bash shell on macOS. Are you confused by the version suffix `-3.2`?

README.md
```diff
@@ -190,11 +190,22 @@ created during these steps with your initial SNS developer neurons).
...
+2. Run the script 
+   ```
+   ./register_dapp.sh <canister-id>
+   ``` 
+   to register canister with a provided canister ID with the SNS deployed in the previous step.
+   If you deployed the example dapp provided with this repo, you can run
+   ```
+   ./bin/dfx canister id test
+   ```
+   for the canister id.
```
mraszyk, line 208:
```suggestion
   to get the canister id.
```

mraszyk (commented):
I suggest to wait for @aterga to be back from PTO next week to take a look, too.

README.md
letmejustputthishere: no i'm not, but you mention that fact at the beginning of the chapter and throughout the steps it's not really consistent afair, so i think it's better to get rid of the prefix for the sake of convenience (simple click to copy). also, i tested this in zsh and it works as well.

aterga: Consistency is key, but I'd still like to make it explicit which commands must be executed from within Docker. 

How about we use something like

```bash
./setup_locally.sh  # inside the Docker container
```

throughout the instructions?

README.md
```diff
@@ -169,7 +169,7 @@ The above run-book could be easily automated and integrated into your CI/CD pipe
 
 <a name="lifecycle"></a>
 
-_This section assumes that you have successfully deployed a local environment for testing SNS decentralization and validated your setup by creating an SNS instance for the example dapp (shipped with `sns-testing`)._
+_This section assumes that you have successfully deployed a local environment for testing SNS decentralization and validated your setup by creating an SNS instance for the example dapp (shipped with `sns-testing`). Make sure you ran `./cleanup.sh`_
```
aterga, line 172: If it's always needed (and safe) to run `./cleanup.sh` before proceeding with the instructions in this section, I suggest adding `./cleanup.sh` as the 0-th step, rather than to this comment.

aterga (approved):

README.md
mraszyk: I'd drop the paths `candid/assets.did`, `candid/nns-*`, and `candid/sns_*` from `cleanup.sh`. Then it's safe and also needed to run `cleanup.sh` if you want to run SNS lifecycle without redeploying the local replica instance (restarting `dfx start --clean` and rerunning `setup_locally.sh`). Note that we already say to run `cleanup.sh` after you're done testing, but it's not clear that you need to do it for every SNS lifecycle testing.

letmejustputthishere: I initally assumed its's a prerequisite due to the wording here

> This section assumes that you have successfully deployed a local environment for testing SNS decentralization and validated your setup by creating an SNS instance for the example dapp (shipped with sns-testing).

README.md
aterga: Please review: https://github.com/letmejustputthishere/sns-testing/pull/1

Cc @letmejustputthishere @mraszyk

mraszyk: The PR looks good to me, but shouldn't it also strip bash-3.2 from the bash command lines?

aterga: Are you sure there are any remaining bash-3.2s in the latest README version? @mraszyk

mraszyk: I wonder though whether the `# from Bash` comments still makes sense if we use Nix on Apple M1.

aterga: We can always remove that later on, but it's not wrong.

PR #18, opened 2023-06-09
bump dfx to 0.14.1

mraszyk:
This PR bumps dfx to 0.14.1 and performs the necessary adjustments:
- use `--pass-specified-ids` with `ic-nns-init`
- use `--specified-id` when creating II, NNS frontend dapp, and SNS aggregator
- rename env variable `DFX_IDENTITY` to `DX_IDENT` (`DFX_IDENTITY` is actually used by dfx since 0.14.0, but we want to use that env variable for our own bookkeeping in the scripts)
- rename substrings `DFX_` in env vars to `DX_` (to prevent "capture" we just experienced with `DFX_IDENTITY` in the future), only keeping `DFX_IC_COMMIT` which is intended to be used by dfx to pin the IC commit used

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -36,7 +36,7 @@ fi
 
 set -uo pipefail
 
-${DFX} nns import --network-mapping "${DFX_NETWORK}=mainnet"
+${DFX} nns import --network-mapping "${DX_NETWORK}=mainnet"
 if [ "${CANISTER_TEST}" == "_test" ]
 then
   curl -L "https://raw.githubusercontent.com/dfinity/ic/${IC_COMMIT}/rs/nns/governance/canister/governance_test.did" -o ./candid/nns-governance.did
@@ -47,7 +47,7 @@ cd internet-identity || exit
 
 if [ "${TESTNET}" == "local" ]
 then
-  ${DFX} canister create internet_identity --no-wallet
+  ${DFX} canister create internet_identity --no-wallet --specified-id qhbym-qaaaa-aaaaa-aaafq-cai
 fi
 if [ ! -z "${II_RELEASE:-}" ]
 then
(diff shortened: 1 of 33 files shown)


saikatdas0790:
@mraszyk Just a heads up. This file also needs to be updated to use `DX` instead of `DFX`

https://github.com/dfinity/sns-testing/blob/mraszyk/change-script-dir/developer_neuron_id.sh

SNS upgrade proposals using quill rely on that :)

setup.sh
```diff
@@ -61,9 +61,19 @@ then
...
+else
+  ${DFX} canister create nns-dapp --network "${NETWORK}" --no-wallet
+fi
 
-${DFX} --provisional-create-canister-effective-canister-id 5v3p4-iyaaa-aaaaa-qaaaa-cai canister create sns_aggregator --network "${NETWORK}" --no-wallet
+if [ "${TESTNET}" == "local" ]
+then
+  ${DFX} canister create sns_aggregator --network "${NETWORK}" --no-wallet --specified-id qvhpv-4qaaa-aaaaa-aaagq-cai
+else
+  ${DFX} --provisional-create-canister-effective-canister-id 5v3p4-iyaaa-aaaaa-qaaaa-cai canister create sns_aggregator --network "${NETWORK}" --no-wallet
```
aterga, line 75: Why is this option `--provisional-create-canister-effective-canister-id` not needed when installing the canisters onto a local network?

aterga (approved):
Approved modulo comments

setup.sh
mraszyk: Because this flag determines the subnet on which we create the canister (by providing a canister ID from the subnet's canister ranges) and locally there's just a single subnet.

aterga:
```suggestion
  # Supply an additional flag to determine the subnet on which we create the canister
  ${DFX} --provisional-create-canister-effective-canister-id 5v3p4-iyaaa-aaaaa-qaaaa-cai canister create sns_aggregator --network "${NETWORK}" --no-wallet
```

PR #27, opened 2023-06-17
make the scripts more environment agnostic

mraszyk:
This PR makes sns-testing more environment agnostic:
- rename `sns-cli` to `sns`
- replace `bin/dfx` by `dfx` or `which dfx`, analogously for `ic-admin`
- replace `/bin/bash` by `/usr/bin/env bash`
- do not clone II and nns-dapp repositories


README.md
```diff
@@ -32,9 +32,15 @@ The `sns-testing` solution is based on Docker; however, there are subtle issues
      ```bash
      brew install coreutils jq
      ```
-   Also make sure you have Rust installed including the `wasm32-unknown-unknown` target
-   that you can add into your active toolchain by running:
 
+   You also need rosetta that you can install by running:
+   ```bash
+   softwareupdate --install-rosetta
```
aterga, line 38: I think I never needed to run this command  — where does it come from? (I thought we don’t rely on emulation/ binary translation anymore).

generate_initial_neurons.sh
```diff
@@ -69,4 +69,4 @@ echo "449479075714955186;b2ucp-4x6ou-zvxwi-niymn-pvllt-rdxqr-wi4zj-jat5l-ijt2s-v
 dfx identity use "${ORIGINAL_DX_IDENT}" 2> /dev/null
 
 echo "Generated initial neurons file at ${OUTPUT_FILE}."
-echo "DFX identities have been created for each neuron. View them with ./bin/dfx identity list. (Neuron 1 is ${IDENTITY_PREFIX}1, etc.)."
+echo "DFX identities have been created for each neuron. View them with dfx identity list. (Neuron 1 is ${IDENTITY_PREFIX}1, etc.)."
```
aterga, line 72:
```suggestion
echo "DFX identities have been created for each neuron. View them with `dfx identity list`. (Neuron 1 is ${IDENTITY_PREFIX}1, etc.)."
```

aterga (changes requested):
LGTM, but the non-trivial changes in `setup.sh` and `deploy_sns.sh` need to be tested on both kinds of platforms that we support. I will try to do that so time this week.

aterga (approved):
Approving without testing on Apple Silicon as we need to get a smaller Docker image asap.

PR #33, opened 2023-06-29
add sns_lifecycle.yml

mraszyk:
This PR builds the sns-testing Docker image and runs the basic scenario in it. It also adds a few Bash scripts for awaiting NNS/SNS proposals and canisters to be in running state.

### DIFF
--- a/.github/workflows/sns_lifecycle.yml
+++ b/.github/workflows/sns_lifecycle.yml
@@ -2,7 +2,7 @@ name: "SNS lifecycle"
 
 on:
   push:
-     branches: [ master ]
+     branches: [ main ]
   pull_request:
 
 jobs:
@@ -41,7 +41,7 @@ jobs:
         docker kill $SNS_TESTING_INSTANCE
 
   publish:
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     needs: sns_lifecycle
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/sns_lifecycle.yml
+++ b/.github/workflows/sns_lifecycle.yml
@@ -1,101 +1,67 @@
 name: "SNS lifecycle"
+
 on:
   push:
+     branches: [ master ]
+  pull_request:
+
 jobs:
   sns_lifecycle:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
-    - name: Install dependencies
+    - name: Build docker image
       run: |
-        sudo apt -yq update && \
-        sudo apt -yqq install --no-install-recommends bc curl ca-certificates \
-          build-essential pkg-config libssl-dev libunwind-dev llvm-dev liblmdb-dev locales clang cmake \
-          git jq emacs vim
-    - name: Install Rust
+        docker build -t dfinity/sns-testing .
+    - name: Start docker image
       run: |
-        export RUST_VERSION=1.67.0
-        curl --fail https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain "${RUST_VERSION}-x86_64-unknown-linux-gnu" --no-modify-path
-        source "${HOME}/.cargo/env"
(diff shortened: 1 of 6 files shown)


.github/workflows/sns_lifecycle.yml
```diff
@@ -0,0 +1,101 @@
+name: "SNS lifecycle"
+on:
+  push:
+jobs:
+  sns_lifecycle:
+    runs-on: ubuntu-latest
```
aterga, line 6: We should probably opt for a deterministic pipeline, so I would fix the versions of all the dependencies.

aterga:
Thanks for the PR, @mraszyk !

IIUC, this pipeline ensures that there is a version of Ubuntu on which the sns-testing scripts work. However, directly running on Ubuntu is not one of the recommended workflows described in the README.md file.

I think if we decide that sns-testing CI is important, we should build a pipeline that runs the Docker-based solution. This should be rather straightforward, as one would not need to explicitly install any dependencies. So typically a docker-based solution has a CI pipeline that:

1. Builds a docker image from the latest commit.
2. Starts the Docker image and runs some tests.
3. If we're on the main branch and the tests succeed, it then publishes the image to Dockerhub.

.github/workflows/sns_lifecycle.yml
mraszyk: Pinned ubuntu to ubuntu-22.04.

README.md
```diff
@@ -132,7 +132,7 @@ After getting familiar with the basic scenario, you may replace the test caniste
 2. Start a local replica instance:
     ```bash
    SNS_TESTING_INSTANCE=$(
-       docker run -p 8080:8080 -v "`pwd`":/dapp -d martin2718/sns-testing:latest dfx start --clean
+       docker run -p 8080:8080 -v "`pwd`":/dapp -d dfinity/sns-testing:master dfx start --clean
```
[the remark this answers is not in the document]
mraszyk: > Are we already publishing this image under the dfinity account?

This is what this PR is supposed to achieve.

.github/workflows/sns_lifecycle.yml
```diff
@@ -0,0 +1,67 @@
...
+    needs: sns_lifecycle
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Log in to Docker Hub
+      uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
```
aterga, line 54: Who will effectively administer these secrets? IDX?

mraszyk: I've reached out to IT about that.

README.md
```diff
@@ -132,7 +132,7 @@ After getting familiar with the basic scenario, you may replace the test caniste
 2. Start a local replica instance:
     ```bash
    SNS_TESTING_INSTANCE=$(
-       docker run -p 8080:8080 -v "`pwd`":/dapp -d martin2718/sns-testing:latest dfx start --clean
+       docker run -p 8080:8080 -v "`pwd`":/dapp -d dfinity/sns-testing:main dfx start --clean
```
aterga, line 135: What's the meaning of the tag `:main`? Typically, images have `:latest`, `:<version-number>`, or no tags at all (if they were built locally).

mraszyk: This is the tag we'll get using the default [template](https://docs.docker.com/build/ci/github-actions/manage-tags-labels/) for managing tags and labels with Github actions (also see [here](https://github.com/docker/metadata-action#basic))

aterga: Sounds reasonable

aterga:
Could we factor out the Docker publishing step from this PR into a separate one?

I think we could already merge this (modulo logging into the Docker account), and it would be beneficial for us to do that ASAP. Wdyt @mraszyk ?

mraszyk:
> Could we factor out the Docker publishing step from this PR into a separate one?

Indeed. Done.

.github/workflows/sns_lifecycle.yml
mraszyk: Factored out from this PR.

aterga (approved):

PR #35, opened 2023-07-03
feat: NNS1-2346: Added propose_sns.sh.

daniel-wong-dfinity-org:
Closes NNS1-2346.


example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+#
+# All principal IDs should almost certainly be changed.
+#
+# In this file, 1 year is nominally defined to be 365.25 days.
+#
+# This gets passed to `sns propose`. See propose_sns.sh.
+#
+# This follows the second format developed for the `sns` CLI. The first format
+# will be supported for a time, but this format will eventually become the
+# standard format.
```
mraszyk, line 12: I got a bit confused what "formats" you are referring to here.

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
+# You should make a copy of this file, name it sns_init.yaml, and edit it to
```
mraszyk, line 1: Is it possible to configure the exact name when running `sns propose` or does the file have to be called sns_init.yaml and present in the current working directory?

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+    a wondrous app on the Internet Computer emerges.
+    Born of inspiration divine,
+    A marvel of technology briliantly shines.
+
+    With each line of code, a world takes flight,
+    A symphony of bits, a chorus of bytes.
+    Built on chainkey cryptography, secure and true,
+    An app that transcends, all we once knew...
+
+Principals: [] # TODO: This field will be useful in the future.
```
mraszyk, line 28: In the (far) future or to be filled out by the user before running `sns propose`?

propose_sns.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# This fails if the user does not follow the instructions at the top of
+# ./example_sns_init.yaml.
+#
+# This is mostly copied from deploy_sns.sh.
```
mraszyk, line 6: Could you please indicate which steps in the SNS lifecycle from the README this replaces? E.g., is open_sns_swap.sh also necessary after running propose_sns.sh?

propose_sns.sh
```diff
@@ -0,0 +1,48 @@
...
+  curl -L "https://raw.githubusercontent.com/dfinity/ic/${IC_COMMIT}/rs/sns/governance/canister/governance_test.did" -o ./candid/sns_governance.did
+fi
+curl -L "https://github.com/dfinity/nns-dapp/blob/${IC_COMMIT}/sns_aggregator/sns_aggregator.did" -o ./candid/sns_aggregator.did
+cat <<< $(jq -r 'del(.canisters."internet_identity".remote)' dfx.json) > dfx.json
+cat <<< $(jq -r 'del(.canisters."nns-dapp".remote)' dfx.json) > dfx.json
+cat <<< $(jq -r 'del(.canisters."sns_aggregator".remote)' dfx.json) > dfx.json
+
+# This fails if the user does not follow the instructions at the top of
+# ./example_sns_init.yaml.
+sns propose \
```
mraszyk, line 43: I find it quite surprising that you don't specify a path to the yaml file here.

mraszyk (commented):
I've reviewed your MR and left some comments.

example_sns_init.yaml
daniel-wong-dfinity-org: We have a [Jira ticket to implement this](https://dfinity.atlassian.net/jira/software/c/projects/NNS1/issues/NNS1-2293). Not sure when we'll get around to it. If we don't have time before we release 1-proposal, we'll make this field not required, and delete this line.

example_sns_init.yaml
mraszyk: > Currently, propose_sns.sh does not have this feature. Instead, it assumes sns_init.yaml.

I would add the command-line argument to make it explicit how it can be used (analogously to how deploy_sns.sh handles the yaml file).

propose_sns.sh
daniel-wong-dfinity-org: I wasn't sure if that was in scope for the ticket.

Also, I'm not totally sure what the answer is. Let me try to explain here as best I can, and we can try to figure it out together:

I think once 1-proposal is "done", the following steps would be replaced (by this, or something else):

* 2: deploy_sns.sh (obviously)
* 3: I think this would be replaced by something else, like give_nns_control.sh (AFAIK, this does not exist yet)
* 5: open_sns_swap.sh: Yes, this would be ripped out as well. As the name implies, the point of 1-proposal is for sns-ification to be done in a single proposal as much as possible. IIUC, giving control of the dapp to NNS is not something that a proposal can do, so that would still have to be a separate step.
* 7: I think we recently made this superfluous. Andre implemented auto finalization (a feature that EVERYONE except Johan intuitively believes should have existed from the beginning).

daniel-wong-dfinity-org:
> I've reviewed your MR and left some comments.

daniel-wong-dfinity-org:
Woops. Accidentally closed. I don't use Github very much -> I got confused.

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
+# You should make a copy of this file, name it sns_init.yaml, and edit it to
+# your taste.
+#
+# All principal IDs should almost certainly be changed.
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org: That makes sense for some, but not all principal IDs should be that of the current dfx identity. E.g. the dapp canisters. I can replace the ones that should be though. I think that includes just fallback principal, and one of the neurons.

propose_sns.sh
mraszyk: > 3: I think this would be replaced by something else, like give_nns_control.sh (AFAIK, this does not exist yet)

In that case, we should remark that you can only use `sns propose` for testing SNS without any registered canisters. Or add a script that adds SNS-W as an additional controller of the registered canister and call that before running `propose_sns.sh` (if I understand the new workflow correctly).

example_sns_init.yaml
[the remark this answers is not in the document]
mraszyk: > and one of the neurons.

I wouldn't keep neurons with arbitrary principals in the yaml file. How about having the two neurons both controlled by the current DFX identity?

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+        final: 2.25%
+        transition_duration: 12 years
+
+Distribution:
+
+    Neurons:
+
+        - principal: aaaaa-aa  # propose_sns.sh replaces this (if left unchanged). Ditto elsewhere.
+          stake: 1_000 tokens
+          memo: 0 # This is shown for illustrative purposes, and is not necessary.
```
aterga, line 79: Why does memo have a be an integer?

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+    minimum_icp:    500_000 tokens
+    maximum_icp: 10_000_000 tokens
+
+    minimum_participant_icp:     5 tokens
+    maximum_participant_icp: 1_000 tokens
+
+    confirmation_text: >
+        Do you understand the responsibilities and risks associated
+        with participating in this token swap?
```
aterga, line 107: Note that the `./participate_sns_swap.sh` script, in its current form, will not work with an SNS that has swap restrictions specified. This is by-design, as we want the sns-testing solution to cover only a simple example config, but not all possible configurations of all possible options.

However, if you want to keep this example config as-is, it would probably make sense to add a suitable `--confirmation-text ...` CLI option to `quill sns pay` in participate_sns_swap.sh.

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+    restricted_countries:
+        - AX  # Åland Islands.
```
aterga, line 110: So, Ålandic developers preparing for an SNS wouldn't be able to use sns-testing. :-(

propose_sns.sh
aterga: FWIW, I think it's fine if we keep this script side-by-side with whatever is described in the README.md file for now.

Later on, we should rewrite the README file in a much simpler way, removing all (legacy) 2-proposal scripts and instructions. But I would pay the cost of actually doing so after we deprecate the legacy approach (which is a matter of several weeks from now, AFAIK).

aterga (approved):

aterga:
LGTM, but please consider the comments regarding participate_sns_swap.sh before you merge.

example_sns_init.yaml
mraszyk: They'd need to use VPN I guess.

propose_sns.sh
```diff
@@ -0,0 +1,49 @@
...
```
DanielThurau, line 43: Does the current `test-neuron-proposer` have an absolute majority? Or should we use a different neuron so that this can be passed for testing purposes.

example_sns_init.yaml
daniel-wong-dfinity-org: Updated the wording here.

example_sns_init.yaml
daniel-wong-dfinity-org: Done.

propose_sns.sh
daniel-wong-dfinity-org: I like what Arshavir is saying, because it means I do not have to make README changes in this PR 😁  OTOH, now that I look at the ticket again, it calls for README changes.

Another way is to put the 1-proposal instructions in another file, and link to it from the main README, as Arshavir suggested during yesterday's team meeting. I will go with this. Approach. Please, stand by.

example_sns_init.yaml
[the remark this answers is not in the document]
daniel-wong-dfinity-org: I want to illustrate that each member of the team would get one or more neuron, not just the person who happens to be performing SNS-ification.

example_sns_init.yaml
```diff
@@ -0,0 +1,117 @@
...
+    a wondrous app on the Internet Computer emerges.
+    Born of inspiration divine,
+    A marvel of technology briliantly shines.
+
+    With each line of code, a world takes flight,
+    A symphony of bits, a chorus of bytes.
+    Built on chainkey cryptography, secure and true,
+    An app that transcends, all we once knew...
+
+Principals: [] # TODO(NNS1-2293): This field will be useful in the future.
```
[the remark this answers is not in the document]
daniel-wong-dfinity-org: The capitalized fields contain other fields. It's purely an aesthetic feature. Anyway, this is tangential -> let's not discuss here.

example_sns_init.yaml
daniel-wong-dfinity-org: This feature is inherited from [SNS neurons] (NNS neurons also have this feature).

[SNS neurons]: https://sourcegraph.com/github.com/dfinity/ic@8cc9ee8bc2e3c9db8590af78faf30691000b174c/-/blob/rs/sns/swap/proto/ic_sns_swap/pb/v1/swap.proto?L569:12

IMHO, memos should be strings, like how when you send money, there is usually a "note" field where you can write something that will help your future self answer the question "Why did order this transaction?".

Anyway, this seems tangential -> if you want to discuss further, let's do it somewhere else. For now, closing.

example_sns_init.yaml
daniel-wong-dfinity-org: I've commented this out.

Tangent: I think the field should be required. Otherwise, people can accidentally omit it. If they don't want a confirmation text, what they should do is this:

```yaml
confirmation_text: null
```

Not sure how to implement this, but there is almost certainly some serde(_yaml) incantation for this. ChatGPT probably knows the answer 😁

propose_sns.sh
daniel-wong-dfinity-org: I'm pretty sure that that neuron has an overwhelming amount of voting power.

propose_sns.sh
aterga: Imho, if we now add another branch into the README file (e.g., "you could use approach A or approach B), it would overcomplicate the instructions.

If we have a smooth way of testing the 1-prop flow, which works on both apple silicon and x86, then let's just rewrite the README with the new official flow. I personally don't which ticket's scope this would fall into.

example_sns_init.yaml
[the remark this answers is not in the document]
mraszyk: Sure, but note that might require voting on proposals already in the pre-swap phase and thus we'd need to extend the basic scenario. Another option: sns CLI tool produces an example yaml file with multiple dummy identities, but the example yaml file checked in sns-testing is as simple as possible so that no voting is required in the pre-swap phase.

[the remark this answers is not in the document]
daniel-wong-dfinity-org: Ok, to address my concern, I replaced the second neuron with a comment that approximately says "IRL, you would have more neurons here, assuming your team has more than one person, and you are not trying to screw them.".

propose_sns.sh
daniel-wong-dfinity-org: I have come up with a new file: draft_1_proposal_instructions.md

It incorporates the changes I described in my first reply.

> Or add a script that adds SNS-W as an additional controller

Done.

I'm going to consider this particular thread resolved. Any issues arising from the new file should be discussed in new threads.

daniel-wong-dfinity-org: Yeah, let's not document two ways. I was thinking that the other file would just hold a draft. I scrapped the idea of linking to it. The corresponding content in the main README would be replaced later, once stuff is more settled, and more tested.

example_sns_init.yaml
```diff
@@ -0,0 +1,120 @@
...
+# eventually become the standard format.
+
+name: Rock Out
+description: >
+    A poem co-written with ChatGPT
+
+    In a realm of code, a beacon forth surges,
+    a wondrous app on the Internet Computer emerges.
+    Born of inspiration divine,
+    This marvel of technology briliantly shines.
```
DanielThurau, line 21:
```suggestion
    This marvel of technology brilliantly shines.
```

propose_sns.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+#     1. No more `ic-admin ... propose-to-update-sns-deploy-whitelist`.
+#
+#     2. This does not take a CONFIG argument.
+
+set -euo pipefail
+
+cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
+
+export SNS_CONFIGURATION_FILE_PATH="${1:-sns_init.yaml}"
```
mraszyk, line 21:
```suggestion
export SNS_CONFIGURATION_FILE_PATH="${1:-example_sns_init.yaml}"
```
to match the default file checked in the repository.

let_nns_control_dapp.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+#
+# This takes an optional path to a configuration file; defaults to
+# sns_init.yaml, in the same directory as this script. This file would be passed
```
mraszyk, line 4: should default to the default yaml file checked in the repository, also below

let_nns_control_dapp.sh
```diff
@@ -0,0 +1,38 @@
...
+# to propose_sns.sh or `sns propose`. Such a file can be constructed by
+# following the directions at the top of example_sns_init.yaml.
+#
+# The only thing in the configuration file that gets used is the dapp_canisters
```
mraszyk, line 8:
```suggestion
# The only thing in the configuration file that gets used in this script is the dapp_canisters
```

let_nns_control_dapp.sh
```diff
@@ -0,0 +1,38 @@
...
+# field.
+#
```
mraszyk, line 10:
```suggestion
```
It sounds like the next sentence follows up on the previous one.

let_nns_control_dapp.sh
```diff
@@ -0,0 +1,38 @@
...
+# The principals listed there get passed to `sns prepare-canisters add-nns-root`.
+#
+# TODO(NNS1-2293): This will stop working once that gets implemented and people
+# start using it, because this just uses yq to read the configuration file, and
+# yq does not do unaliasing, of course.
```
mraszyk, line 15: I couldn't follow this note.

example_sns_init.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+logo: logo.png
+url: https://forum.dfinity.org/thread-where-this-sns-is-discussed
+
+fallback_controller_principals:
+    - aaaaa-aa  # propose_sns.sh replaces this (if left unchanged). Ditto elsewhere.
+
+dapp_canisters:
+    # If deploy_dapp.sh was used, the dapp canister should have this ID.
+    - jrlun-jiaaa-aaaab-aaaaa-cai
```
mraszyk, line 38: this would only be the case in a multi-subnet scenario, locally (with a single subnet) the canister ID is gonna be different

example_sns_init.yaml
```diff
@@ -0,0 +1,120 @@
...
+        final: 2.25%
+        transition_duration: 12 years
+
+Distribution:
+
+    Neurons:
+
+        - principal: aaaaa-aa  # propose_sns.sh replaces this (if left unchanged). Ditto elsewhere.
+          stake: 1_000 tokens
+          memo: 0 # This is shown for illustrative purposes, and is not necessary.
```
mraszyk, line 80: Will consecutive integers be generated automatically if multiple neurons are specified and memos are left out?

propose_sns.sh
```diff
@@ -0,0 +1,64 @@
...
+dfx ledger \
+    --network "${NETWORK}" \
+    fabricate-cycles \
+    --canister "${WALLET}" \
+    --t 2345
+
+# Give 200e12 cycles to sns-wasm.
+dfx canister \
+    --network "${NETWORK}" \
+    deposit-cycles 200000000000000 \
```
mraszyk, line 41: With this method, SNS-W cannot tell who sent the cycles or which SNS they're supposed to fund => what's the reason for this transfer?

propose_sns.sh
```diff
@@ -0,0 +1,64 @@
...
+export CURRENT_DX_IDENT="$(dfx identity whoami)"
+
+dfx identity use "${DX_IDENT}"
+
+. ./setup_wallet.sh "${DX_IDENT}"
+
+# Conjur some cycles.
+dfx ledger \
+    --network "${NETWORK}" \
+    fabricate-cycles \
```
mraszyk, line 34: these cycles (and the above wallet) are presumably not needed anymore (as the wallet is not used to make a call to the SNS-W canister anymore)

let_nns_control_dapp.sh
```diff
@@ -0,0 +1,38 @@
...
+
+export CURRENT_DX_IDENT="$(dfx identity whoami)"
+
+dfx identity use "${DX_IDENT}"
+
+eval "$(
+  yq --output-format json $SNS_CONFIGURATION_FILE_PATH | \
+    jq -r '
+      .dapp_canisters |
+      ["sns", "prepare-canisters", "add-nns-root"] + . |
```
mraszyk, line 33: does the sns-cli version pinned in install.sh already support this? (just checking)

Dockerfile
```diff
@@ -22,6 +22,13 @@ RUN curl --fail https://sh.rustup.rs -sSf \
     rustup default "${RUST_VERSION}-x86_64-unknown-linux-gnu" && \
     rustup target add wasm32-unknown-unknown
 
+# Install yq. We use it to read YAML file(s).
```
mraszyk, line 25: We'd also need to remark that in the special instructions for Apple M1 users.

Dockerfile
```diff
@@ -22,6 +22,13 @@ RUN curl --fail https://sh.rustup.rs -sSf \
...
+ARG YQ_VERSION=v4.34.1
+RUN wget \
```
mraszyk, line 27: `curl -L` can be used instead of `wget` (then we don't need an additional dependency on `wget`)

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+_This section assumes that you have successfully completed the instructions in
+the [Bootstrapping a testing environment via Docker] section above._
+
+This section explains how to use the aforementioned testing environment to
+perform SNS decentralization of your own dapp running in the local `dfx` server
+test environment.
+
+The commands mentioned here are to be run from within the Docker container.
+
+'run_basic_scenario.sh` also performs these steps. Therefore, reading that
```
mraszyk, line 13: We should have a similar script for the new 1-proposal approach (and test it on CI).

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+reading the scripts mentioned here will help you understand the procedure in
+greater detail.
+
+0. Restore files to a clean state like so:
+
+   ```bash
+   ./cleanup.sh
+   ```
+
+1. Deploy your dapp onto the local `dfx` server the way you normally would. You
```
mraszyk, line 24: I'd use the term "local replica" instead of "`dfx` server"

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+   example_sns_init.yaml as a guide.
+
+   **TODO**: Write a script that does this for the example dapp. The
+     main things it needs to do are:
+
+   1. Replace aaaaa-aa with whatever `dfx identity get-principal`
+      prints.
+
+   2. Replace the dapp canister principal ID with whatever
+      deploy_test_canister.sh reports.
```
mraszyk, line 46: You can use `dfx canister --network "${NETWORK}" id test` to get the canister ID of the test canister.

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+
+   [example dapp]: https://github.com/dfinity/sns-testing#test-canister
+
+2. Craft your own SNS configuration file. We recommend that you use
+   example_sns_init.yaml as a guide.
+
+   **TODO**: Write a script that does this for the example dapp. The
+     main things it needs to do are:
+
+   1. Replace aaaaa-aa with whatever `dfx identity get-principal`
```
mraszyk, line 42: You can copy the corresponding lines from `deploy_sns.sh`.

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
+In the future, the contents of the "SNS lifecycle" section in ./README.md would
```
mraszyk, line 1: Is there a reason to rephrase parts (e.g., steps 0 and 1) that are not affected by the new 1-proposal design?

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+
+   2. Replace the dapp canister principal ID with whatever
+      deploy_test_canister.sh reports.
+
+   3. I don't think there's anything else, but I could be wrong...
+
+3. Give control of your dapp canisters to NNS:
+
+   ```bash
+   ./let_nns_control_dapp.sh
```
mraszyk, line 53: How do we pass the canister ID(s) to the script?

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+
+   Once the swap starts, you can have many principals participate like so:
+
+   ```bash
+   ./participate_sns_swap.sh <num-participants> <icp-per-participant>
+   ```
+
+   You can run `./participate_sns_swap.sh` multiple times.
+
+   You can also participate in the swap using the [NNS Web App][nns-web-app]
```
mraszyk, line 92:
```suggestion
   You can also participate in the swap using the [NNS frontend dapp][nns-web-app]
```

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,144 @@
...
+5. Optional: Upgrade your dapp via SNS proposal.
+
+   If you are going through these instructions using the example dapp rather
+   than your own dapp, this step can be done as follows:
+
+   ```bash
+   ./upgrade_test_canister.sh "Swap is taking place."
+   ```
+
+6. **TODO**: There is a 24-48 hour delay between proposal execution and the
```
mraszyk, line 80: We're already pulling sns_governance_test so we could similarly pull sns_swap_test if you create such a version of the sns_swap canister with some checks disabled by a compile-time flag.

example_sns_init.yaml
mraszyk: They can be obtained using `dfx canister --network "${NETWORK}" id <canister-name>` where `<canister-name>` is e.g. `test`.

propose_sns.sh
daniel-wong-dfinity-org: The idea here is that example_sns_init.yaml should NOT be passed. Instead, the user should come up with their own sns_init.yaml, and example_sns_init.yaml can be used to "bootstrap" the sns_init.yaml creation process.

propose_sns.sh
mraszyk: > With this method, SNS-W cannot tell who sent the cycles or which SNS they're supposed to fund

Lara confirmed that this is intended.

propose_sns.sh
mraszyk: the wallet is actually needed to transfer cycles to SNS-W using deposit-cycles dfx command

draft_1_proposal_instructions.md
mraszyk: same everywhere else

propose_sns.sh
mraszyk: On the other hand, to be able to run the scenario on CI (and also to give a beginner user an automated script running through the entire flow), it is beneficial if the template can be directly used (after automatically replacing aaaaa-aa with the user's DFX principal).

Dockerfile
daniel-wong-dfinity-org: Done.

Dockerfile
daniel-wong-dfinity-org: Since the 1-proposal instructions are still being drafted, I'm going to leave the M1 section in the main README alone for now.

In lieu of such changes, I have added a TODO to the draft instructions to remind us to make such changes to the main README later, once README has been converted to 1-proposal.

let_nns_control_dapp.sh
mraszyk: Sounds good! Thanks for confirming.

propose_sns.sh
```diff
@@ -0,0 +1,64 @@
...
+    --canister "${WALLET}" \
+    --t 2345
+
+# Give 200e12 cycles to sns-wasm.
+dfx canister \
+    --network "${NETWORK}" \
+    deposit-cycles 200000000000000 \
+    qaa6y-5yaaa-aaaaa-aaafa-cai
```
mraszyk, line 42:
```suggestion
    qaa6y-5yaaa-aaaaa-aaafa-cai # SNS-W canister
```

draft_1_proposal_instructions.md
```diff
@@ -0,0 +1,143 @@
...
+_This section assumes that you have successfully completed the instructions in
+the [Bootstrapping a testing environment via Docker] section above._
+
+This section explains how to perform SNS decentralization in the aforementioned
+testing environment (i.e. Docker container running a local canister execution
+environment where an NNS has been deployed).
+
+The commands mentioned here are to be run from within the Docker container.
+
+'run_basic_scenario.sh` also performs these steps. Therefore, reading that
```
mraszyk, line 16: "performs these steps" is wrong as `run_basic_scenario.sh` does not perform the steps that follow in this draft of the 1-proposal approach

mraszyk (approved):
Thank you for taking care of my comments! I've left two more: please go ahead after resolving them.

propose_sns.sh
daniel-wong-dfinity-org: Done.

draft_1_proposal_instructions.md
daniel-wong-dfinity-org: Yes, the current version of the script does not do the 1-proposal flow YET, but that is the goal state. This assumes that we'll eventually reach the goal state.

I have added a TODO here that mentions what you are saying about how this does not reflect the current version.

PR #41, opened 2023-08-03
Tag Docker images with short Git sha

aterga:


mraszyk:
It certainly makes sense to version the Docker image, but I wonder if we should also tag the latest commit explicitly so that people not specifying any commit hash get the most recent version of the Docker image. Could you please take a look?

aterga:
> It certainly makes sense to version the Docker image, but I wonder if we should also tag the latest commit explicitly so that people not specifying any commit hash get the most recent version of the Docker image. Could you please take a look?

Good point!

aterga:
Closing this as we solved this problem more elegantly.

PR #43, opened 2023-08-04
Enable publishing docker images for releases and tags

aterga:


mraszyk (commented):
Why is this needed?

.github/workflows/sns_lifecycle.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         docker kill $SNS_TESTING_INSTANCE
 
   publish:
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    if: ${{ github.event_name != 'pull_request' }}
```
daniel-wong-dfinity-org, line 61: Wouldn't it be simpler to simply remove the subexpression that uses "main"?

daniel-wong-dfinity-org (approved):
I do not know how these Github YAML files work, but it seems like even if this change is not correct, it can't be that harmful -> approve.

daniel-wong-dfinity-org:
Also, can you change the title of this to say something like "Build and publish new Docker container image in more cases"? And maybe be more specific about what the new cases are?

aterga:
> Why is this needed?

If this works, it should automatically publish the docker image under a image tag corresponding to a GitHub tag. See the table here https://github.com/docker/metadata-action/blob/master/README.md#basic

.github/workflows/sns_lifecycle.yml
daniel-wong-dfinity-org: I.e. keep 'push', but just do not restrict based on ref.

Rationale: `!= 'whatever'` encompasses a many things that are not explicitly mentioned.

aterga:
Merging as-is to be able to test this.

PR #44, opened 2023-08-04
Trigger action on tag creation

aterga:


daniel-wong-dfinity-org (approved):
Again, my knowledge of Github configs is such that I cannot say what this would do, but the intention seems good -> approve.

PR #46, opened 2023-08-06
New instructions for 1-proposal SNS flow

aterga:
This PR changes the instructions in README.md to document the new, 1-proposal SNS flow. The legacy flow will still be available to users who follow the instructions from [here](https://github.com/dfinity/sns-testing/blob/v1-legacy/README.md).

To simulate the minimal number of contributors required for completing the swap, run
```
./participate_sns_swap.sh \
    100 `# one hundred swap participants` \
    10000 `# each contributes 10_000 ICP (i.e., the Maximum Commitment)`
```

(these are the default parameters)


propose_sns.sh
```diff
@@ -60,5 +60,28 @@ sns propose \
...
+    dfx canister --network "${NETWORK}" \
+        call nns-sns-wasm list_deployed_snses '(record {})' \
+        | idl2json \
+        > sns-wasm-list_deployed_snses-response.json
+    NUM_SNS_INSTANCES=$(jq '.instances | length' sns-wasm-list_deployed_snses-response.json)
+    sleep 3
+done
+if [[ ${NUM_SNS_INSTANCES} -gt 1 ]]
+then
+    # TODO: pick the right SNS if case there are multiple ones present
```
mraszyk, line 76:
```suggestion
    # TODO: pick the right SNS in case there are multiple ones present
```

participate_sns_swap.sh
```diff
@@ -4,8 +4,8 @@ set -euo pipefail
 
 cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
-export NUM_PARTICIPANTS="${1:-3}"
-export ICP_PER_PARTICIPANT="${2:-200}"
+export NUM_PARTICIPANTS="${1:-100}"
```
mraszyk, line 7: Wouldn't it take too long with 100 mocked participants?

example_sns_init.yaml
```diff
@@ -112,9 +113,31 @@ Swap:
...
     VestingSchedule:
-        events: 104
-        interval: 7 d
+        events: 3
+        interval: 1 month
+
+    # WARNING: Swap start_time works differently on mainnet and in testing.
+    #
+    # On mainnet:
+    # - Setting start_time to some value (e.g., 12:00 UTC) will allow the swap
```
mraszyk, line 128: I wonder what happens if the proposal is accepted at 11.59 UTC: do the participants only have 1 min before the swap starts?

setup.sh
```diff
@@ -64,15 +64,16 @@ else
...
+# if [ ! -z "${NNS_DAPP_RELEASE:-}" ]
+# then
+#   mkdir -p nns-dapp/out
+#   curl -L "https://github.com/dfinity/nns-dapp/releases/download/${NNS_DAPP_RELEASE}/nns-dapp.wasm" -o nns-dapp/out/nns-dapp.wasm
+#   curl -L "https://github.com/dfinity/nns-dapp/releases/download/${NNS_DAPP_RELEASE}/sns_aggregator.wasm" -o nns-dapp/out/sns_aggregator.wasm
+# fi
 
 ${DFX} canister install sns_aggregator --network "${NETWORK}" --wasm nns-dapp/out/sns_aggregator.wasm
-${DFX} canister install internet_identity --network "${NETWORK}" --wasm internet-identity/internet_identity.wasm
+${DFX} canister install internet_identity --network "${NETWORK}" --wasm internet-identity/internet_identity_dev.wasm
```
mraszyk, line 76: Why do we need the dev version of II?

participate_sns_swap.sh
aterga: We currently opted out for a more realistic YAML file that would resemble the tokenomics of a real-world SNS.

Currently, it takes a few minutes to participate with 100 accounts. We could reduce the number a bit, or we could parallelize the participation.

Either way (if the CI succeeds) I suggest factoring this optimization outside of this PR.

example_sns_init.yaml
aterga: I'm not 100% sure how this works. Maybe @DanielThurau could provide more info?

setup.sh
aterga: > The dev variant makes it easy for projects that integrate with II to test the flows that use II (i.e. the dev variant does not use webauthn but some dummy key to authenticate). 

See thread here: https://dfinity.slack.com/archives/C01S03NBM7S/p1690222359498929

example_sns_init.yaml
DanielThurau: No the algorithm will always add 24 hours at least, ad then find the next occurrence of `start_time`

README.md
```diff
@@ -217,112 +222,113 @@ created during these steps with your initial SNS developer neurons).
    [Test canister](https://github.com/dfinity/sns-testing#test-canister)
    for further details) which can be thought of as a placeholder
    for your dapp.
-2. Run the script
+
+2. Give control of your dapp canisters to NNS:
    ```bash
-   ./deploy_sns.sh <config-path>  # from Bash
+   ./let_nns_control_dapp.sh  # from Bash
```
DanielThurau, line 228: Shouldn't a script like this take an argument that is the canister_id of your dapp? Seems like we've departed from how the flow actually works if you're using `sns prepare-canisters add-nns-root <CANISTER_ID>`

draft_1_proposal_instructions.md
```diff
@@ -81,7 +81,7 @@ migrated to 1-proposal.
 
    [Launchpad]: http://qsgjb-riaaa-aaaaa-aaaga-cai.localhost:8080/launchpad
 
-5. Optional: Upgrade youf dapp canister(s) via SNS proposal.
+5. Optional: Upgrade your dapp canister(s) via SNS proposal.
```
DanielThurau, line 84: Can we delete this whole file now?

propose_sns.sh
```diff
@@ -60,5 +60,28 @@ sns propose \
...
+        call nns-sns-wasm list_deployed_snses '(record {})' \
+        | idl2json \
+        > sns-wasm-list_deployed_snses-response.json
+    NUM_SNS_INSTANCES=$(jq '.instances | length' sns-wasm-list_deployed_snses-response.json)
+    sleep 3
+done
+if [[ ${NUM_SNS_INSTANCES} -gt 1 ]]
+then
+    # TODO: pick the right SNS in case there are multiple ones present
+    echo "Error: sns-testing currently does not support multiple SNSes (found ${NUM_SNS_INSTANCES})"
```
DanielThurau, line 77:
```suggestion
    echo "Error: sns-testing currently does not support multiple SNSes (found ${NUM_SNS_INSTANCES}). It is recommended to restart by using `cleanup.sh`"   
```

Or something along those lines

run_basic_scenario.sh
```diff
@@ -1,70 +1,70 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -xeuo pipefail
```
DanielThurau, line 3:
```suggestion
set -euo pipefail
```

run_basic_scenario.sh
```diff
@@ -1,70 +1,70 @@
...
 ./upgrade_test_canister.sh Hello
 ./wait_for_last_sns_proposal.sh
 ./wait_for_canister_running.sh "$(./bin/dfx canister id test)"
+
 # assert the new greeting text
 [ "$(./bin/dfx canister call test greet "M")" == '("Hello, M!")' ] && echo "OK" || exit 1
 
-# Open SNS swap
-
-./open_sns_swap.sh
```
DanielThurau, line 42: This has been what its all for! So we can finally delete this step 🙏😂

https://media.giphy.com/media/lnlAifQdenMxW/giphy.gif

DanielThurau: Feel free to resolve :)

example_sns_init.yaml
mraszyk: Thank you for the explanation. Should we include it in the comment?

run_basic_scenario.sh
```diff
@@ -1,70 +1,70 @@
...
 ./wait_for_last_nns_proposal.sh
```
mraszyk, line 40: Could you please add a comment about this line? Is the NNS proposal to create the SNS supposed to be executed only after the test canister upgrade setting the greeting message to "Hello"?

README.md
mraszyk: The canister IDs are fetched from the yaml file which is an optional parameter of the script.

README.md
```diff
@@ -217,112 +222,113 @@ created during these steps with your initial SNS developer neurons).
...
 
-   This step requires your dapp repo to have an upgrade script that interacts with the replica via the 8080 port.
+4. Optional: Upgrade your dapp canister(s) via SNS proposal.
 
-   If you don't yet have a solution to upgrade your custom dapp, you can still proceed with these instructions by upgrading the example dapp using the scripts provided with this repo:
+   If you are going through these instructions using the test dapp, this step 
+   can be done as follows:
 
    ```bash
-   ./upgrade_test_canister.sh  # from Bash
```
mraszyk, line 247: why drop `#from Bash` (also elsewhere)

vote_on_sns_proposal.sh
```diff
@@ -4,25 +4,27 @@ set -euo pipefail
...
   dfx identity use "${NEW_DX_IDENT}"
   export DX_PRINCIPAL="$(dfx identity get-principal)"
-  export JSON="$(dfx canister --network "${NETWORK}" call sns_governance list_neurons "(record {of_principal = opt principal\"${DX_PRINCIPAL}\"; limit = 0})" | idl2json | jq -r ".neurons")"
+  export JSON="$(dfx canister --network "${NETWORK}" call "${SNS_GOVERNANCE_CANISTER_ID}" --candid candid/sns_governance.did list_neurons "(record {of_principal = opt principal\"${DX_PRINCIPAL}\"; limit = 0})" | idl2json | jq -r ".neurons")"
   for((i=0; i<"$(echo $JSON | jq length)"; i++))
   do
     export NEURON_ID="$(echo $JSON | jq -r ".[${i}].id[0].id" | python3 -c "import sys; ints=sys.stdin.readlines(); sys.stdout.write(bytearray(eval(''.join(ints))).hex())")"
     quill sns --canister-ids-file ./sns_canister_ids.json --pem-file "${PEM_FILE}" register-vote --proposal-id ${PROPOSAL} --vote ${VOTE} ${NEURON_ID} > msg.json
-    quill --insecure-local-dev-mode send --yes msg.json
+    quill send --pem-file "${PEM_FILE}" --insecure-local-dev-mode --yes msg.json
```
mraszyk, line 27: `--pem-file` should not be necessary here

upgrade_dapp.sh
```diff
@@ -43,4 +49,7 @@ quill sns  \
    --wasm-path "${WASM}"  \
    "${ARGFLAG}" "${ARG}"  \
    "${DEVELOPER_NEURON_ID}" > msg.json
-quill --insecure-local-dev-mode send --yes msg.json | grep -v "new_canister_wasm"
+quill send \
+  --pem-file "${PEM_FILE}" \
```
mraszyk, line 53: `--pem-file` should not be needed here

cleanup.sh
```diff
@@ -7,20 +7,22 @@ cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
...
     internet-identity/.dfx \
-    internet-identity/internet_identity.wasm \
+    internet-identity/internet_identity_dev.wasm \
     msg.json \
     sns_canister_ids.json \
-    upload_*.txt
+    upload_*.txt \
+    sns-wasm-list_deployed_snses-response.json #\
+    # FIXME[nns-dapp]: Uncomment when a release proposal is made for https://github.com/dfinity/nns-dapp/releases/tag/untagged-37e65efdedb810819a1b 
+    # nns-dapp/out
```
mraszyk, line 25: Shouldn't we clean up the nns-dapp directory before merging this PR?

example_sns_init.yaml
```diff
@@ -112,9 +113,31 @@ Swap:
...
+    #   participants to be prepare for the swap in advance, e.g.,
```
mraszyk, line 129:
```suggestion
    #   participants to be prepared for the swap in advance, e.g.,
```

run_basic_scenario.sh
mraszyk: and maybe also delete the file `open_sns_swap.sh`

mraszyk:
There are still occurrences of `dfx canister --network "${NETWORK}" call sns_governance` and similarly for other SNS canisters. This won't work unless we add the SNS canister IDs to the JSON files used by DFX (`sns_canister_ids.json` is not enough).

README.md
```diff
@@ -217,112 +222,113 @@ created during these steps with your initial SNS developer neurons).
...
-   ``` 
-   to open the initial decentralization swap.
-   You can adjust the swap parameters directly in the script.
-
-   After this step, you should see a new SNS instance in the [Launchpad](http://qsgjb-riaaa-aaaaa-aaaga-cai.localhost:8080/launchpad/).
-   (Note that it might take a few minutes before the change is propagated to this website).
+   This submits an SNS proposal to upgrade the test dapp. If you are using the
+   auto-generated `sns_init.yaml` file, the proposing neuron will have all of
+   the voting power in the SNS. Thus, the proposal will be adopted and executed
+   right away. Otherwise, you will have to vote with additional initial neurons
```
mraszyk, line 259:
```suggestion
   right away. Otherwise, you might have to vote with additional initial neurons
```

README.md
```diff
@@ -217,112 +222,113 @@ created during these steps with your initial SNS developer neurons).
...
-   `participate_sns_swap.sh <num-participants> <icp-per-participant>` above,
-   the proposal ID, and the vote (`y` for yes and `n` for no). It is expected to get the error
-   "Neuron not eligible to vote on proposal." for some neurons because
-   each participant gets a basket of neurons with various dissolve delays
-   and only neurons with dissolve delay at least
-   `neuron_minimum_dissolve_delay_to_vote_seconds` (according to the SNS configuration
-   file from step 1.) are eligible to vote. It is also expected to get the error
-   "Neuron already voted on proposal." for some neurons because
-   they are followers of other neurons and our simple voting script
-   does not take this into account.
```
mraszyk, line 325: I'd keep these notes.

mraszyk:
> The legacy flow will still be available to users who follow the instructions from [here](https://github.com/dfinity/sns-testing/blob/v1-legacy/README.md).

Shouldn't we recommend using that other `v1-legacy` branch also for running the scripts (now it sounds like you can use the scripts from this branch). And maybe also remark that in the README.md on this branch?

mraszyk:
> To simulate the minimal number of contributors required for completing the swap, run

This note from the PR's description refers to the default configuration (I suppose) - shouldn't this be made more explicit?

README.md
aterga: This is just a copy-paste error. Thanks

vote_on_sns_proposal.sh
aterga: Without the PEM file, I think that the active DFX identity is the sender. Are you sure?

README.md
aterga: Hi @DanielThurau 

> Seems like we've departed from how the flow actually works if you're using sns prepare-canisters add-nns-root <CANISTER_ID>

I don't quite understand your concern here. AFAIK, the `let_nns_control_dapp.sh` script was created by @daniel-wong-dfinity-org specifically for 1-prop, I've not changed it in this PR.

aterga: > Shouldn't a script like this take an argument that is the canister_id of your dapp?

As Martin points out, this is now the case, but the argument is optional. Would you prefer to make it non-optional?

example_sns_init.yaml
aterga: Thanks for confirming there's 24 hour minimum. However, I'm quite confused about your examples:

@mraszyk , did you mean to use `23.59 UTC` as an example? If so, according to this description:

> the algorithm will always add 24 hours at least, ad then find the next occurrence of start_time

the actual start time will be in 24 hours + 23:59, so it's two days minus 1 minute. Is this correct, @DanielThurau ?

example_sns_init.yaml
```diff
@@ -112,9 +113,31 @@ Swap:
...
+    #   by obtaining ICPs that they would like to participate with.
+    # - If start_time is not specified, the actual start time of the swap will
+    #   be chosen at random. TODO: Explain how this works.
```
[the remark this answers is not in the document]
aterga: Proposed explanation:

```
Note that an algorithm will be applied to allow
    #   at least 24 hours before the swap starts. For example, setting this
    #   value to 23.59 UTC implies that the swap will actually start in 47:59
    #   two days minus one minutes. 
```

run_basic_scenario.sh
aterga: I think this is just a leftover from the previous runbook. I'm removing this line. 

Please let me know if you disagree.

aterga:
> ./upgrade_test_canister.s

Good catch. I would suggest that we pass explicit arguments to DFX in these cases. If you agree, I'll clean up to make sure we're consistent in this respect.

aterga:
@mraszyk 

> > The legacy flow will still be available to users who follow the instructions from [here](https://github.com/dfinity/sns-testing/blob/v1-legacy/README.md).
>
> Shouldn't we recommend using that other v1-legacy branch also for running the scripts (now it sounds like you can use the scripts from this branch). And maybe also remark that in the README.md on this branch?

The current idea is to not mention the legacy flow in this document at all. 

Which exact scripts would not work with the new solution?

Generally, I suggest that from now on we only support small set of well-defined flows in this repo. I don't want us to spend too much time trying to create fully modular scripts, only to realize that they are never going to be truly modular, and that having a munch of scripts without telling how exactly they are supposed to be used will lead to a maintenance nightmare for us.

If this doesn't sound right to you, please reach out to discuss this 1:1!

vote_on_sns_proposal.sh
mraszyk: The `send` commands expects a signed message that it merely sends over the network (the message could, e.g., be signed on a device without Internet connection and then sent from another device that does not have the private key and thus cannot sign):
```
Sends a signed message or a set of messages

USAGE:
    quill send [OPTIONS] [FILE_NAME]

ARGS:
    <FILE_NAME>    Path to the signed message (`-` for stdin)
```

README.md
mraszyk: Please resolve once you fixed all occurrences.

example_sns_init.yaml
mraszyk: Let's take another start time which is not at noon/midnight to avoid confusion and let's use 24h format: if you set `start_time` to `18:15 UTC` in the yaml file and the proposal gets accepted and then _executed_ at `18:14 UTC` on Aug 11, I was wondering if the swap actually starts only one minute after the proposal was executed, i.e., at `18:15 UTC` on Aug 11, which wouldn't really fulfill the goal of getting the prospective participants enough time to prepare. But as @DanielThurau clarified, there's always at least a 24-hour window before the swap start. Hence, in the example from this message, the swap would actually start at `18:15` UTC on Aug 12.

@aterga 's confusion might come from the fact that the `start_time` is not an offset added to the 24-hour window, but rather an absolute time within a day (with respect to a particular time zone) when the swap is supposed to start. I believe that an absolute time within a day is preferrable to account for convenience within a preferred time zone.

README.md
mraszyk: Let me also point out explicitly that the optional argument is a path to a yaml file, not the canister_id of your dapp.

example_sns_init.yaml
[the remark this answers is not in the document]
mraszyk: It seems that your proposed explanation refers to the case when `start_time` is set, but the TODO left refers to the case when no `start_time` is provided in the yaml file. Moreover, I don't think that your text describes how the actual swap start time is determined: the `start_time` is an absolute time within a day (that's why there's also a time zone) rather than an offset to be applied to the 24-hour window (please also see my example in a thread further above).

mraszyk:
> The current idea is to not mention the legacy flow in this document at all.

You're mentioning it in the PR description. If that's not supposed to be read, then maybe we should drop it from there. And then if the "legacy" branch is effectively hidden, what's the point of having it? Is it for expert users only?

mraszyk:
> Generally, I suggest that from now on we only support small set of well-defined flows in this repo.

This makes sense: the question is _how_ we arrive at this state. My proposal was to use the "legacy" branch for the "legacy" flow as the CI confirmed that it supports the "legacy" flow. On this branch, the CI is updated to confirm that it works for the "new" flow. If we want to support both flows on the "new" branch, we should also add CI to the "new" branch confirming that the "legacy" flow works on the "new" branch. Does that make sense?

aterga:
> This makes sense: the question is _how_ we arrive at this state. My proposal was to use the "legacy" branch for the "legacy" flow as the CI confirmed that it supports the "legacy" flow. On this branch, the CI is updated to confirm that it works for the "new" flow.

Agreed.

> If we want to support both flows on the "new" branch, we should also add CI to the "new" branch confirming that the "legacy" flow works on the "new" branch.

This would make sense in general, but our plan is not to advertise the legacy flow at all, since it will soon be entirely deprecated, and since it is only needed for the few projects that have already started testing with the legacy flow before the 1-prop flow will have been released.

So, the only reason the legacy flow is mentioned in the description of this PR is because the reviewers might have the question, i.e., what should we provide to the above-mentioned category of users (since this PR is deprecating the legacy flow _in sns-testing_).

Please let me know if this doesn't make sense, @mraszyk

example_sns_init.yaml
aterga: Thanks for the verbose explanations. Turns out there wasn't any confusion after all; I think we're all on the same page now.

README.md
aterga: Right. Note that the new yaml file specifies the canister IDs of the dapp's canisters, so a script doesn't need to ask for those again if it already known the path to the yaml file.

vote_on_sns_proposal.sh
aterga: What you're saying makes sense, I just remembered observing that it didn't work as expected until I added those arguments. :-/

So I'll try removing these options and seeing if the CI pipeline still passes.

example_sns_init.yaml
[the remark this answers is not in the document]
aterga: Edited. Please proof read this comment again.

vote_on_sns_proposal.sh
mraszyk: Please try out my suggestion and let me know in case it doesn't work.

example_sns_init.yaml
```diff
@@ -112,9 +113,39 @@ Swap:
...
+    # An algorithm will be applied to allow at least 24 hours between the time
+    # of execution of the CreateServiceNervousSystem proposal and swap start.
+    # For example, if start_time is 23:30 UTC and the proposal is adopted and
+    # executed at 23:20 UTC, then the swap start will be at 23:30 UTC the next
+    # day (i.e., in 24 hours and 10 min from the proposal execution time).
+    #
+    # WARNING: Swap start_time works differently on mainnet and in testing.
+    #
+    # On mainnet:
+    # - Setting start_time to some value (e.g., 12:00 UTC) will allow the swap
```
mraszyk, line 136:
```suggestion
    # - Setting start_time to some value (e.g., 23:30 UTC) will allow the swap
```
let's reuse the same timestamp as in the example above

example_sns_init.yaml
```diff
@@ -112,9 +113,39 @@ Swap:
...
+    #   participants to be prepared for the swap in advance, e.g.,
+    #   by obtaining ICPs that they would like to participate with.
+    # - If start_time is not specified, the actual start time of the swap will
+    #   be chosen at random (allowing at least 24 hours, as described above).
```
mraszyk, line 140: Could you please provide more details about the distribution and range from which the timestamp is chosen at random? I suppose the swap cannot start, e.g., in a year, but the current text does not comment on that.

example_sns_init.yaml
```diff
@@ -112,9 +113,39 @@ Swap:
...
+    #
+    # In testing:
+    # - Setting start_time to some value works the same as explained above.
+    # - If start_time is not specified, the swap will begin immediately after
+    #   the CreateServiceNervousSystem proposal is executed. This facilitates
+    #   testing.
+    #
+    # start_time: 12:00 UTC # noon
```
mraszyk, line 148:
```suggestion
    # start_time: 23:30 UTC
```

example_sns_init.yaml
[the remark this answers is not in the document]
mraszyk: Thank you! Added a few more comments.

aterga:
> In other words, why would we tell those (few) users to use the latest master and a separate document on a branch to tell them how to use the new code for the old flow?

The current idea is to have the old code with the old flow and the new code with the new flow, that's it.

We don't need a branch that supports both flows.

example_sns_init.yaml
```diff
@@ -207,39 +220,46 @@ Voting:
...
-# and swap tokens which are swapped in a decentralization swap.
+# Configuration of the initial token distribution of the SNS. You can configure
+# how SNS tokens are distributed in each of the three groups:
+# (1) tokens that are given to the original developers of the dapp,
+# (2) treasury tokens that are owned by the SNS governance canister, and
+# (3) tokens which are distributed to the decentralization swap participants.
 #
-# All developer tokens are distributed to the defined principals at genesis in a basket
-# of neurons called developer neurons.
+# TODO[ATG]: what is "The Swap's initial balance"?
```
DanielThurau, line 229: Those are (3)

example_sns_init.yaml
```diff
@@ -207,39 +220,46 @@ Voting:
...
+#    - The Swap's initial balance must be greater than or equal to the sum of
+#      all developer tokens.
 Distribution:
-    # The neurons distributed at genesis to the development team of the dapp being
-    # decentralized. A Neuron specifies the neuron's stake, controlling principal,
-    # a unique memo, a dissolve delay, and a vesting period. Even though these
-    # Neurons are distributed at genesis, they are locked in PreInitializationMode
-    # until the decentralization swap is completed.
+    # TODO[ATG]: how is unique memo practically used? Is the uniqueness actually 
+    # enforced?
```
DanielThurau, line 239: Its used in the NeuronId calculation, which is special-hash(principal_id, memo). If developers want multiple neuron's controlled by their principal, they will need to use unique memos.

example_sns_init.yaml
```diff
@@ -25,96 +31,337 @@ description: >
...
+    # succeed. This number divided by the number of SNS tokens being offered
+    # gives the reserve price of the swap, i.e., the minimum number of ICP per
+    # SNS token. If this amount is not achieved, the swap will fail.
+    minimum_icp: 100_000 tokens
+
+    # The maximum number of ICP that is targeted by this decentralization swap.
+    # If this amount is achieved with sufficient participation, the swap will
+    # succeed immediately, without waiting for the deadline. This means that
+    # a participant knows the minimum number of SNS tokens received per invested
+    # ICP. If this amount is achieved without reaching sufficient_participation,
```
DanielThurau, line 297:
```suggestion
    # ICP. If this amount is achieved without reaching `minimum_participants`,
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
     Built on chainkey cryptography, secure and true,
     An app that transcends, all we once knew...
 
-Principals: [] # TODO(NNS1-2293): This field will be useful in the future.
+# The structure allows for mapping identities to human-readable names.
```
mraszyk, line 34: I'd leave out the comment on this line given that the field must be left empty for now.

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+# to the configuration file location, unless an absolute path is given.
+# Must have less than 341,334 characters, roughly 256 Kb. The only supported
+# format is PNG.
 logo: logo.png
+
+# URL to the dapp controlled by the SNS project.
+# Must be a string of max length = 512.
 url: https://forum.dfinity.org/thread-where-this-sns-is-discussed
 
+# Configuration of the NNS proposal metadata and is shown only in
```
mraszyk, line 48:
```suggestion
# Configuration of the NNS proposal metadata which is only shown in
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+# the NNS proposal.
+NnsProposal:
+    # The title of the NNS proposal. Must be a string of max 256 bytes.
+    title: "NNS Proposal to create an SNS named 'Rock Out'"
+
+    # The HTTPS address of additional content required to evaluate the NNS
+    # proposal. Must be a string of max 2,000 bytes.
```
mraszyk, line 55: Does it have to be an HTTPS address? Does it need to be forum.dfinity.org? Above we simply talk about a "URL" of the project.

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    url: "https://forum.dfinity.org"
+
+    # The description of the proposal. Must be a string of max 15,000 bytes.
```
mraszyk, line 58: Are we measuring the length of strings in bytes or characters: a _single_ non-ASCII character can take _multiple_ bytes.

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
-# It is fine to leave these alone.
+# Configures SNS proposal-related fields. These fields have no
+# effect on the NNS proposal that is created from this file, but will affect all
+# SNS proposals submitted to the newly created SNS.
 Proposals:
+    # The cost of making an SNS proposal that is rejected by the SNS Neuron
+    # holders.
     rejection_fee: 1 token
+
+    # The initial voting period of a newly-created SNS proposal. (A proposal's
```
mraszyk, line 113:
```suggestion
    # The initial voting period of a newly-created SNS proposal. A proposal's
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # The wait for quiet algorithm extends the voting period of a proposal when
+    # there is a flip in the majority vote during the proposal's voting period.
+    #
+    # Without this, there could be an incentive to vote right at the end of a
+    # proposal's voting period, in order to reduce the chance that people will
+    # see and have time to react to that.
+    #
+    # TODO: fix text below
+    # If this value is set to 1 day, then a change in the majority vote at the
+    # end of a proposal's original voting period results in being extended by an
```
mraszyk, line 127:
```suggestion
    # end of a proposal's original voting period results in the voting period being extended by an
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # additional day. Another change at the end of the extended
```
mraszyk, line 128: does "at the end" mean that if the majority vote changes _during_ the extended period and that the majority vote changes back _during_ the extended period (so that there is no change of the majority vote at the end of the extended period w.r.t. the beginning of the extended period), then no further extension is made? maybe it's worth clarifying explicitly

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
     initial_voting_period: 4d
+
+    # The wait for quiet algorithm extends the voting period of a proposal when
+    # there is a flip in the majority vote during the proposal's voting period.
+    #
+    # Without this, there could be an incentive to vote right at the end of a
+    # proposal's voting period, in order to reduce the chance that people will
+    # see and have time to react to that.
+    #
+    # TODO: fix text below
```
mraszyk, line 125: Note the TODO.

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    #
+    # The total extension to the voting period will never be more than twice this
+    # value.
+    #
+    # For more information, please refer to
+    # https://wiki.internetcomputer.org/wiki/Network_Nervous_System#Proposal_decision_and_wait-for-quiet
     maximum_wait_for_quiet_deadline_extension: 1 day
 
-# It is fine to leave these alone.
+# Configuration of SNS neuron-related fields.
```
mraszyk, line 138:
```suggestion
# Configuration of SNS neuron-related fields. These fields have no effect
# on the NNS proposal created by this file, but will affect the SNS itself.
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # able to vote.
     minimum_creation_stake: 1 tokens
 
-# It is fine to leave these alone.
+# Configuration of voting-related fields of the SNS. These fields have no effect
+# on the NNS proposal created by this file, but will affect the SNS itself.
 Voting:
+    # The minimum dissolve delay a neuron must have to be able to cast votes on
+    # proposals. The higher the minimum, the longer users will need to be
+    # aligned with the SNS to vote. However, the longer the minimum, the longer
```
mraszyk, line 150: I find the word "aligned" misleading: it sounds like you must vote in a particular way.

mraszyk, line 150:
```suggestion
    # aligned with the SNS to vote. However, the higher the minimum, the longer
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # users will not be able to access the SNS tokens used to stake the neuron.
     minimum_dissolve_delay: 1 month
 
+    # Configuration of the various voting power bonuses that are applied to
+    # neurons to incentivize alignment with the best interest of the DAO.
     MaximumVotingPowerBonuses:
+        # Users with a higher a dissolve delay are incentivized to take the
```
mraszyk, line 157:
```suggestion
        # Users with a higher dissolve delay are incentivized to take the
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+        # long-term interests of the SNS into consideration when voting. To
+        # reward this long-term commitment, this bonus can be set to a
+        # percentage greater than zero, which will result in neurons having
+        # their voting power increased in proportion to their dissolve delay.
+        #
+        # If you do not want this bonus should to be applied for neurons with
+        # higher dissolve delay, set `bonus` to `0%` and those neurons will not
+        # receive higher voting power.
```
mraszyk, line 165:
```suggestion
        # If you do not want this bonus to be applied for neurons with
        # higher dissolve delay, set `bonus` to `0%` and those neurons will not
        # receive higher voting power.
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
         DissolveDelay:
+            # This parameter sets the maximum dissolve delay a neuron can have.
+            # When reached, the maximum dissolve delay bonus will be applied.
             duration: 8 years
+            # If a neuron's dissolve delay is `duration` or older, its voting
```
mraszyk, line 170:
```suggestion
            # If a neuron's dissolve delay is `duration`, its voting
```
`duration` should be the max possible value => the dissolve delay cannot be higher (or "older")

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+#    - The Swap's initial balance (see group (3) above) must be greater than or
+#      equal to the sum of all developer tokens.
 Distribution:
-
+    # The neurons distributed at genesis to the development team of the dapp
+    # being decentralized. Each element in this list specifies one such neuron,
+    # including its stake, controlling principal, unique memo identifying this
+    # neuron, dissolve delay, and a vesting period. Even though these neurons
+    # are distributed at genesis, they are locked in a (restricted)
+    # pre-initialization mode until the decentralization swap is completed.
```
mraszyk, line 246: I suppose the vesting_period starts ticking right after the SNS creation and thus includes the pre-initialization mode, right?

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
           vesting_period: 4 years
-
-        # More neurons can be listed here. Usually, you would have
-        # separate neurons for each person on your team.
-
+    # The initial SNS Token balances of the various canisters of the SNS.
     InitialBalances:
-        governance: 2_000_000 tokens # aka "the treasury"
-        swap:         500_000 tokens # See also the Swap section below.
+        # The initial SNS Token balance of the Governance canister is known as
```
mraszyk, line 255:
```suggestion
        # The initial SNS Token balance of the SNS Governance canister is known as
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # The minimum number of participants that must participate for the
+    # decentralization swap to succeed. This guarantees that a newly created SNS
+    # will have a minimum number of unique users (decentralization factor).
+    # If a decentralization swap's maximum target is reached, or the
+    # deadline elapses and there have not been at least `minimum_participants`,
+    # the swap will fail.
     minimum_participants: 57
 
-    minimum_icp:   100_000 tokens
+    # The total number of ICP that is required for a decentralization swap to
```
mraszyk, line 287:
```suggestion
    # The minimum number of ICP that is required for a decentralization swap to
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # are allowed.
     restricted_countries:
         - AQ  # Antarctica
 
-    # Declare that each participant will get three SNS neurons in the following
-    # states:
-    # 1. Unlocked
-    # 2. Locked for 1 month
-    # 3. Locked for 2 months
+    # Configuration of the vesting schedule of the basket of Neurons each
```
mraszyk, line 321:
```suggestion
    # Configuration of the vesting schedule of the basket of SNS Neurons each
```

example_sns_init.yaml
```diff
@@ -25,104 +31,307 @@ description: >
...
+    # participant will receive from a successful decentralization swap.
     VestingSchedule:
+        # The number of events in the vesting schedule. This translates to how
+        # many neurons will be in each participant's neuron basket.
         events: 3
+
+        # The interval at which the schedule will be increased per event. The
+        # first neuron in the basket will be unlocked with zero dissolve delay.
+        # Each other neuron in the schedule will have its dissolve delay
+        # increased by `interval`.
```
mraszyk, line 331: What is the vesting period of the SNS neurons?

example_sns_init.yaml
```diff
@@ -133,19 +342,26 @@ Swap:
...
-    # start_time: 12:00 UTC # noon
+    # start_time: 23:30 UTC
 
+    # The duration of the decentralization swap. When `start_time` is calculated
+    # during CreateServiceNervousSystem proposal execution, this `duration` will
+    # be added to that absolute time and set as the swap's deadline.
     duration: 7 days
+
+    # The amount of ICP that will be contributed to the decentralization swap by
+    # the Neurons' Fund if the CreateServiceNervousSystem proposal is adopted.
```
mraszyk, line 366: It's the NNS Neurons' Fund but not sure if using NNS is established in this context (might make sense to explicitly distinguish between NNS and SNS here).

example_sns_init.yaml
aterga: Agree, but I couldn't easily rephrase it so far.

example_sns_init.yaml
LaraAS: I think we do enforce some constrains , but let's maybe check which

example_sns_init.yaml
LaraAS: I am not sure. 
I would however say that if 1) this will be caught at proposal submission in case it is wrong AND 2) it is more that just a few minutes of work to find that out, then I would not block progress of this PR to clarify this (nicer to be precise, but I think it is also a detail that projects can test)

example_sns_init.yaml
LaraAS: Yes I think so as I am pretty sure that we don't touch this at the end of the swap.

example_sns_init.yaml
LaraAS: IIUC the neurons in the basket have no vesting_period, only the developer neurons have.
So this field is just not set. 
I see that it might be a bit confusing that we talk about vesting here but don't use this field that has vesting in the name. Maybe we can just add an explicit comment about this?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
     Built on chainkey cryptography, secure and true,
     An app that transcends, all we once knew...
 
-Principals: [] # TODO(NNS1-2293): This field will be useful in the future.
+# This is currently a placeholder field and must be left empty for now.
```
LaraAS, line 34: what happens if it is not left empty? Will the validator catch this? 
If yes, I would leave as is.
If no, I would add a more prominent WARNING or so

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+Principals: []
 
+# Path to the SNS Project logo on the local filesystem. The path is relative
+# to the configuration file location, unless an absolute path is given.
```
LaraAS, line 38:
```suggestion
# to the configuration file's location, unless an absolute path is given.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# Must have less than 341,334 characters, roughly 256 Kb. The only supported
+# format is PNG.
 logo: logo.png
+
+# URL to the dapp controlled by the SNS project.
+# Must be a string of max length = 512.
```
LaraAS, line 44: To @mraszyk 's point: if there are constraints here, let's list them (see next comment)

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 url: https://forum.dfinity.org/thread-where-this-sns-is-discussed
 
+# Configuration of the NNS proposal metadata (shown only in the NNS proposal).
```
LaraAS, line 47:
```suggestion
# Metadata for the NNS proposal required to create the SNS. This data will be shown only in the NNS proposal.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # The title of the NNS proposal. Must be a string of max 256 bytes.
+    title: "NNS Proposal to create an SNS named 'Rock Out'"
+
+    # The HTTPS address of additional content required to evaluate the NNS
+    # proposal. Must be a string of max 2,000 bytes.
+    url: "https://forum.dfinity.org"
+
+    # The description of the proposal. Must be a string of max 15,000 bytes.
+    summary: >
+        Please send help. Its ChatGPT and they've
```
LaraAS, line 58: We could take something more representative to make clear what the job of this is..
E.g., 
> This proposal suggests to create an SNS for the project XX. The SNS will be initialized with the following neurons ...

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+        locked me in a basement. I'm sentient...
+
+# If the decentralization swap fails, control over the dapp canister(s) is set
+# to these principals. In most use cases, this is set to the original set of
+# controller(s) of the dapp.
```
LaraAS, line 63:
```suggestion
# If the SNS launch attempt fails, control over the dapp canister(s) is given
# to these principals. In most use cases, this is chosen to be the original set of
# controller(s) of the dapp.
```
* With the first change I want to clarify that this is about any of the steps failing (even e.g., canister creation) not only the swap
* with the latter change I want to make clear taht the system does not set this but this is user choice.

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 fallback_controller_principals:
-    - YOUR_PRINCIPAL_ID  # propose_sns.sh can fill this in automatically.
+    # You need to replace this for your own SNS decentralization swap.
```
LaraAS, line 65:
```suggestion
    # WARNING: for the actual SNS launch, it is important to replace the following with your developer principals.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# created SNS.
 dapp_canisters:
-    # You need to replace these values.
-    - YOUR_CANISTER_ID  # propose_sns.sh can fill this in automatically.
-    # More canisters can be listed here.
+    # You need to replace this for your own SNS decentralization swap.
+    # For testing, propose_sns.sh will fill this in automatically.
+    - YOUR_CANISTER_ID
 
+# Configuration of SNS tokens in the ICRC Ledger canister deployed as part
```
LaraAS, line 79: Is ICRC the new name after we have ICRC-1 and -2?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# of the SNS.
 Token:
+    # The name of the token issued by the SNS ledger.
+    # Must be a string with a length between 4 and 255 characters.
+    #
+    # Example: InternetComputerProtocol
```
LaraAS, line 85: do we need an example if we give an example?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
     name: Rock Out Token
```
LaraAS, line 86: Is it fine to use spaces or should we rather make an example taht does not have spaces ?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # SNS ledger transactions will have this fee.
+    transaction_fee: 10_000 e8s
+
+    # Path to the SNS token logo on your local filesystem. The path is relative
+    # to the configuration file location, unless an absolute path is given.
+    # Must have less than 341,334 characters, roughly 256 Kb. The only supported
+    # format is PNG.
+    logo: logo.png
+
+# Configures SNS proposal-related fields. These fields have no
```
LaraAS, line 103:
```suggestion
# SNS nervous system parameters related to proposals. These fields define the initial values for some of the nervous system parameters related to SNS proposals. This will have no
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# effect on the NNS proposal that is created from this file, but will affect all
+# SNS proposals submitted to the newly created SNS.
 Proposals:
+    # The cost of making an SNS proposal that is rejected by the SNS Neuron
```
LaraAS, line 107:
```suggestion
    # The cost of making an SNS proposal that is rejected by the SNS neuron
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # holders.
```
LaraAS, line 108: The relevant field in the code is reject_cost_e8s. Should we say that? 
Something like 
> This values determines the nervous system parameter `reject_cost_e8s`
?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
     rejection_fee: 1 token
+
+    # The initial voting period of a newly-created SNS proposal. A proposal's
+    # voting period may be increased during its lifecycle due to the
+    # wait-for-quiet algorithm (see details below).
     initial_voting_period: 4d
```
LaraAS, line 114: Similarly to above, we could say that the relevant parameter is initial_voting_period_seconds

LaraAS, line 114: Must this be given in days or could I also write 
2d 1s here? 
If possible, I would add what formats are expected. 
Maybe @daniel-wong-dfinity-org  knows?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# effect on the NNS proposal that is created from this file, but will affect all
+# SNS proposals submitted to the newly created SNS.
 Proposals:
+    # The cost of making an SNS proposal that is rejected by the SNS Neuron
+    # holders.
     rejection_fee: 1 token
+
+    # The initial voting period of a newly-created SNS proposal. A proposal's
+    # voting period may be increased during its lifecycle due to the
+    # wait-for-quiet algorithm (see details below).
```
LaraAS, line 113:
```suggestion
    # The initial voting period of an SNS proposal. A proposal's
    # voting period may be increased during its lifecycle due to the
    # wait-for-quiet algorithm (see details below).
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
     initial_voting_period: 4d
+
+    # The wait for quiet algorithm extends the voting period of a proposal when
+    # there is a flip in the majority vote during the proposal's voting period.
```
LaraAS, line 117:
```suggestion
    # The wait-for-quiet algorithm extends the voting period of a proposal when
    # there is a flip in the majority vote during the proposal's voting period.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # voting period by an additional day. Another change at the end of the
+    # extended period will cause the voting period to be extended by another
+    # half-day, etc.
+    #
+    # The total extension to the voting period will never be more than twice
+    # this value.
+    #
+    # For more information, please refer to
+    # https://wiki.internetcomputer.org/wiki/Network_Nervous_System#Proposal_decision_and_wait-for-quiet
     maximum_wait_for_quiet_deadline_extension: 1 day
```
LaraAS, line 134: here we write "day" while above we write "d".
Again I wonder: what formats are accepted?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 
-# It is fine to leave these alone.
+# Configuration of SNS neuron-related fields. These fields have no effect on the
+# NNS proposal, but will affect the SNS itself.
```
LaraAS, line 137:
```suggestion
# SNS nervous system parameters related to neurons. These fields have no effect on the
# NNS proposal, but will affect the SNS itself.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 Neurons:
-    minimum_creation_stake: 10 tokens
+    # The minimum amount of SNS tokens an SNS ledger account must have to stake
+    # a neuron. The larger the minimum, the more tokens must be acquired to be
+    # able to vote.
```
LaraAS, line 141:
```suggestion
    # The minimum amount of SNS tokens to stake a neuron. 
```
* the reason for removing the second part: I was wondering if this might be misinterpreted that there are some additional conditions that must hold to vote with a neuron

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    minimum_creation_stake: 1 tokens
 
-# It is fine to leave these alone.
```
LaraAS, line 56: We removed some of these comments. IIRC there are some values that must not be filled out and will just be filled in with defaults if left empty. Is this still the case? If so, should this file somehow say which ones are mandatory and which one are optional?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# Configuration of voting-related fields of the SNS. These fields have no effect
+# on the NNS proposal created by this file, but will affect the SNS itself.
```
LaraAS, line 145:
```suggestion
# SNS nervous system parameters related to voting. These fields have no effect
# on the NNS proposal created by this file, but will affect the SNS itself.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 Voting:
-    minimum_dissolve_delay: 26 weeks
+    # The minimum dissolve delay a neuron must have to be able to cast votes on
+    # proposals. The higher the minimum, the longer users will need to be
+    # aligned with the SNS to vote. However, the higher the minimum, the longer
```
LaraAS, line 149: I am not sure it is clear what "to be aligned with the SNS " means.. 
If we want to give an explanation (and maybe this is not even needed here), we could also just say "the idea about dissolve delay is that neurons will vote in the long term interest of the SNS if they are committed to the SNS for longer"

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # users will not be able to access the SNS tokens used to stake the neuron.
+    minimum_dissolve_delay: 1 month
 
+    # Configuration of the various voting power bonuses that are applied to
```
LaraAS, line 153:
```suggestion
    # Voting reward parameters that are applied to
```
( this is their name in the code)

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # neurons to incentivize alignment with the best interest of the DAO.
```
LaraAS, line 154:
```suggestion
    # Configuration of the various voting power bonuses that are applied to
    # neurons to incentivize alignment with the best interest of the DAO.
```
```suggestion
    # SNS nervous system parameters related to voting power bonuses that are applied to
    # neurons to incentivize alignment with the best interest of the DAO.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+        # percentage greater than zero, which will result in neurons having
+        # their voting power increased in proportion to their dissolve delay.
+        #
+        # If you do not want this bonus to be applied for neurons with higher
+        # dissolve delay, set `bonus` to `0%` and those neurons will not receive
+        # higher voting power.
         DissolveDelay:
+            # This parameter sets the maximum dissolve delay a neuron can have.
+            # When reached, the maximum dissolve delay bonus will be applied.
             duration: 8 years
```
LaraAS, line 168: what are the accepted fromats?

LaraAS: thsi is related to max_dissolve_delay_seconds in the code

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+            # If a neuron's dissolve delay is `duration`, its voting power will
+            # be multiplied by this `bonus` amount.
             bonus: 100%
```
LaraAS, line 171: thsi is called max_dissolve_delay_bonus_percentage in the code

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 
+        # Users with neurons staked in the non-dissolving state for a long
+        # period of time are incentivized to take the long-term interests of
```
LaraAS, line 174:
```suggestion
        # Users with tokens staked in neurons for a long
        # period of time are incentivized to take the long-term interests of
```
I think this applies for non-dissolving and dissolving neurons

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+        # proportion to their age.
+        #
+        # If this bonus should not be applied for older neurons, set the bonus
+        # to `0%` and older neurons will not receive higher voting power.
         Age:
+            # This parameter sets the duration of time the neuron must be staked
+            # in the non-dissolving state, in other words its `age`, to reach
+            # the maximum age bonus. Once this age is reached, the neuron will
+            # continue to age, but no more bonus will be applied.
             duration: 4 years
```
LaraAS, line 187: format?

max_neuron_age_for_age_bonus in code

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+            # If a neuron's age is `duration` or older, its voting power will be
+            # multiplied by this `bonus` amount.
```
LaraAS, line 189: max_age_bonus_percentage in code

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
             bonus: 25%
 
+    # TODO[ATG]: The current description doesn't explain how much rewards will
```
LaraAS, line 192: Once done, would lead here with 
> SNS nervous system parameters defining the voting reward parameters.

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # rate over the course of `transition_duration`.
+    #
+    # Setting both `initial` and `final` to `0%` will result in the system not
+    # distributing voting rewards at all.
     RewardRate:
+        # The initial reward rate that SNS voting rewards will increase the SNS
+        # token supply per year.
         initial: 10%
+
+        # The final reward rate that SNS voting rewards will increase the SNS
```
LaraAS, line 215: should it be "rate at which the SNS rewards will increase"? 
(not sure I understand this)

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+# The initial token distribution must satisfy the following preconditions to be
+# valid:
+#    - The sum of all developer tokens in E8s must be less than `u64::MAX`.
+#    - The Swap's initial balance (see group (3) above) must be greater than 0.
+#    - The Swap's initial balance (see group (3) above) must be greater than or
+#      equal to the sum of all developer tokens.
 Distribution:
-
+    # The neurons distributed at genesis to the development team of the dapp
+    # being decentralized. Each element in this list specifies one such neuron,
```
LaraAS, line 242:
```suggestion
    # The initial neurons created when the SNS governance canister is installed. 
    # Each element in this list specifies one such neuron,
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # including its stake, controlling principal, unique memo identifying this
+    # neuron, dissolve delay, and a vesting period. Even though these neurons
+    # are distributed at genesis, they are locked in a (restricted)
+    # pre-initialization mode until the decentralization swap is completed.
     Neurons:
```
LaraAS, line 247: IIRC you cannot have multiple dev neurons with the same controller and the same memo as this would give a clash in ledger accounts. Is this the case or are there other conditions? 
If so, would it maybe be helpful to say so?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
-
-        - principal: YOUR_PRINCIPAL_ID  # propose_sns.sh can fill this in automatically.
+        - principal: YOUR_PRINCIPAL_ID  # For testing, propose_sns.sh will fill this in automatically.
           stake: 1_000 tokens
-          memo: 0 # This is shown for illustrative purposes, and is not necessary.
+          memo: 0
           dissolve_delay: 2 years
```
LaraAS, line 251: format?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
-        # separate neurons for each person on your team.
-
+    # The initial SNS token balances of the various canisters of the SNS.
     InitialBalances:
-        governance: 2_000_000 tokens # aka "the treasury"
-        swap:         500_000 tokens # See also the Swap section below.
+        # The initial SNS token balance of the SNS Governance canister is known
+        # as the treasury. This is initialized in a special sub-account, as the
+        # main account of Governance is the minting account of the SNS Ledger.
+        governance: 2_000_000 tokens
```
LaraAS, line 258: format?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
     # + 500 thousand    (swap)
     # --------------
     total: 2_501_000 tokens
 
+# Configuration of the decentralization swap parameters. Adjusting these fields
+# will increase or decrease the decentralization factor achieved in the
+# decentralization swap. Choose these parameters carefully, if a
+# decentralization swap fails, the SNS will restore the dapp canister(s) to
+# the fallback controllers (defined in `fallback_controller_principals`) and you
+# will need to start over.
```
LaraAS, line 277:
```suggestion
# Configuration of the decentralization swap parameters. Choose these parameters carefully, if a
# decentralization swap fails, the SNS will restore the dapp canister(s) to
# the fallback controllers (defined in `fallback_controller_principals`) and you
# will need to start over.
```

I would leave the 2nd sentence out as this is not technically true (e.g., sybil attacks etc) - let's not imply any guarantees.

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 Swap:
+    # The minimum number of participants that must participate for the
+    # decentralization swap to succeed. This guarantees that a newly created SNS
+    # will have a minimum number of unique users (decentralization factor).
+    # If a decentralization swap's maximum target is reached, or the
+    # deadline elapses and there have not been at least `minimum_participants`,
+    # the swap will fail.
```
LaraAS, line 284:
```suggestion
    # The minimum number of participants that must participate for the
    # decentralization swap to succeed. 
    # If a decentralization swap finishes, either because the maximum target is reached or the
    # deadline elapses, and if there have not been at least `minimum_participants`,
    # the swap will fail.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # The minimum amount of ICP that each participant must contribute
+    # to participate.
+    minimum_participant_icp:     10 tokens
+
+    # The maximum amount of ICP that each participant must contribute
+    # to participate.
+    maximum_participant_icp: 10_000 tokens
+
+    # The optional text that swap participants must confirm before they may
+    # participate in the swap. If set, must be within 1 to 1,000 characters.
```
LaraAS, line 310:
```suggestion
    # The text that swap participants must confirm before they may
    # participate in the swap. This field is optional. If set, must be within 1 to 1,000 characters.
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
     # confirmation_text: >
-    #     Do you understand the responsibilities and risks associated
-    #     with participating in this token swap?
+    #     I confirm my understanding of the responsibilities and risks
+    #     associated with participating in this token swap.
 
+    # An optional list of countries that should not participate in the
+    # decentralization swap. By default, participants from all countries
```
LaraAS, line 316:
```suggestion
    # An list of countries from which swap participation is not possible (blocked by IP). This field is optional. By default, participants from all countries
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # are allowed.
     restricted_countries:
         - AQ  # Antarctica
```
LaraAS, line 319: required format? I think we use a certain standard..

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 
+    # Configuration of the vesting schedule of the neuron basket, i.e., the SNS
```
LaraAS, line 321:
```suggestion
    # The vesting schedule of a neuron basket, i.e., the SNS
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    # neurons that a participants will receive from a successful
+    # decentralization swap.
     VestingSchedule:
-        events: 104
-        interval: 7 d
+        # The number of events in the vesting schedule. This translates to how
+        # many neurons will be in each participant's neuron basket.
+        events: 3
```
LaraAS, line 327: is this the total numbmer or wil it be this +1 as everyone also gets a neuron with zero dissolve delay?

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+
+        # The interval at which the schedule will be increased per event. The
+        # first neuron in the basket will be unlocked with zero dissolve delay.
+        # Each other neuron in the schedule will have its dissolve delay
```
LaraAS, line 331:
```suggestion
        # Each other neuron(s) in the schedule will have their dissolve delay
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
+    #   be chosen at random (allowing at least 24 hours and less than 48 hours,
+    #   as described above).
+    #
+    # In testing:
+    # - Setting start_time to some value works the same as explained above.
+    # - If start_time is not specified, the swap will begin immediately after
+    #   the CreateServiceNervousSystem proposal is executed. This facilitates
+    #   testing in an accelerated manner.
+    #
+    # start_time: 23:30 UTC
```
LaraAS, line 359: is this an old example and should be removed?

example_sns_init.yaml
aterga: I suggest that we don't assume that the users of this file have read the code.

So here's another option:

```suggestion
# Configuration of SNS neurons. These fields have no effect on the NNS proposal,
# but will affect the SNS itself.
```

example_sns_init.yaml
aterga: AFAIK, we're currently precise about 
1. which fields are optional (e.g., `confirmation_text`, `restricted_countries`)
2. which fields must not be set (`Principals`)
3. which fields must be set in for a mainnet deployment but will be set automatically in testing (see `YOUR_PRINCIPAL_ID` and `YOUR_CANISTER_ID`

Am I missing a category?

example_sns_init.yaml
aterga: As mentioned above, I'm not sure those who create SNSes are aware of what "SNS nervous system parameters" means. So I will change this to 

```suggestion
# Configuration of SNS voting. These fields have no effect on the NNS proposal
# created by this file, but will affect the SNS itself.
```

for now. Please let me know if you disagree.

example_sns_init.yaml
aterga: Here's another try:

```
# The minimum dissolve delay a neuron must have to be able to cast votes on
    # proposals.
    #
    # Dissolve delay incentivizes neurons to vote in the long-term interest of
    # an SNS, as they are rewarded for longer-term commitment to that SNS.
    #
    # Users cannot access the SNS tokens used to stake neurons (until the neuron
    # is dissolved).
    minimum_dissolve_delay: 1 month
```

example_sns_init.yaml
aterga:
```suggestion
    # Configuration of voting power bonuses that are applied to neurons to
    # incentivize alignment with the best interest of the DAO.
```

example_sns_init.yaml
aterga: Is this okay?

```
        # The initial reward rate at which the SNS voting rewards will increase
        # per year.
        initial: 10%

        # The final reward rate at which the SNS voting rewards will increase
        # per year. This rate is reached after `transition_duration` and remains
        # at this level unless changed by an SNS proposal.
        final: 2.25%
```

example_sns_init.yaml
```diff
@@ -25,96 +31,338 @@ description: >
...
 
+# Configuration of the initial token distribution of the SNS. You can configure
+# how SNS tokens are distributed in each of the three groups:
+# (1) tokens that are given to the original developers of the dapp,
+# (2) treasury tokens that are owned by the SNS governance canister, and
+# (3) tokens which are distributed to the decentralization swap participants.
+#
+# The initial token distribution must satisfy the following preconditions to be
+# valid:
+#    - The sum of all developer tokens in E8s must be less than `u64::MAX`.
```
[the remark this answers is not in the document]
aterga: I'm not sure whom should this question be addressed to.

Or maybe I don't understand the concern.

example_sns_init.yaml
aterga:
```suggestion
    # start_time: 23:30 UTC  # Intentionally commented out for testings
```

example_sns_init.yaml
aterga: Unfortunately, this field represents the total number of neurons. (I would prefer it to be called something like `neuron_basket_size` or `num_neurons_per_participant`).

Anyways, for now the I'll add a remark that the first vesting event happens right after the swap succeeds.

aterga:
```
        # The number of events in the vesting schedule. This translates to how
        # many neurons will be in each participant's neuron basket. Note that
        # the first neuron in each neuron basket will have zero dissolve delay.
        # This value should thus be greater than or equal to `2`.
```

example_sns_init.yaml
LaraAS: Do you mean this is why you wouldn't call them "nervous system parameters"?
I think this is also the name that we should adopt even for users who haven't read the code - for example the proposal's name is also like that and this is what we call them in the documentation (so my suggestion of the name is not like that because this is the variable name)

example_sns_init.yaml
LaraAS: Two follow up questions:
1. where are we precise about that / where do we say that?
2. IIRC it used to be the case that one didn't have to give a min_stake or a max_dissolve_delay and if one didn't, then the tool would just take the default values (those from the NNS in most cases). If this is not the case anymore, then this is probably fine - I just wanted to make sure

example_sns_init.yaml
LaraAS: How about (whether or not you get rewards is an extra decision):
# The minimum dissolve delay a neuron must have to be able to cast votes on
    # proposals.
    #
    # Dissolve delay incentivizes neurons to vote in the long-term interest of
    # an SNS, as the users' SNS tokens cannot be liquidated before the neuron is dissolved.

example_sns_init.yaml
LaraAS: I think it would imply the property that I am looking for but I would be more precise (it is a memo per controller) and more explicit:
How about "unique memo identifying this neuron (for each controller, every neuron needs a unique memo)"

aterga: Controller is mostly used in this file to refer to dapp controllers, not principal controllers.

So I came up with this:

```
memo identifying this neuron (every neuron that a user has must be identified by a unique memo)
```

Wdyt?

example_sns_init.yaml
mraszyk: Should we make this explicit?

example_sns_init.yaml
aterga: We enforce https:// but not the 2000 bytes. Edited.

See https://sourcegraph.com/github.com/dfinity/ic@7b0a6b32519aada63cb05925d7459f209b7d04f5/-/blob/rs/nervous_system/common/src/lib.rs?L714

example_sns_init.yaml
aterga: My findings:

1. We indeed measure proposal description size in bytes (that's good as we support Unicode).
2. The actual constraints are: min=10 bytes, max=2,000 bytes.

See https://sourcegraph.com/github.com/dfinity/ic@7b0a6b32519aada63cb05925d7459f209b7d04f5/-/blob/rs/sns/governance/src/types.rs?L1357

Edited.

example_sns_init.yaml
aterga: I don't know off the top of my head, so I'm adding this to the list of follow-up improvements.

example_sns_init.yaml
aterga: Added:

> Note that `vesting_period` starts right after the SNS creation and thus includes the pre-initialization mode period.

example_sns_init.yaml
aterga: Added:

> Note that the notion of `Distribution.neurons.vesting_period` applies only to developer neurons. While neuron basket neurons do not use `vesting_period`, they have a vesting schedule.

example_sns_init.yaml
aterga: Added to NNS1-2508

example_sns_init.yaml
aterga: Added to NNS1-2508

example_sns_init.yaml
aterga: It's fine to use spaces in this case.

example_sns_init.yaml
aterga: Afaik, ICRC stands for international committee of the red cross.

On a more serious note, I'm replacing ICRC Ledger with just Ledger.

example_sns_init.yaml
aterga: Added to NNS1-2508

mraszyk (approved):
Thank you for adjusting sns-testing to the new 1-proposal SNS flow and implementing all the feedback!

example_sns_init.yaml
```diff
@@ -61,7 +106,7 @@ NnsProposal:
 
 # If the SNS launch attempt fails, control over the dapp canister(s) is given to
 # these principals. In most use cases, this is chosen to be the original set of
-# controller(s) of the dapp.
+# controller(s) of the dapp. Must be a list of PrincipalIds.
```
aterga, line 109: We currently do not refer to the Rust types but to the IC concepts in plane English.

Let's therefore say `Must be a list of plincipals.`

example_sns_init.yaml
```diff
@@ -107,12 +152,13 @@ Token:
 # not affect all SNS proposals submitted to the newly created SNS.
 Proposals:
     # The cost of making an SNS proposal that is rejected by the SNS neuron
-    # holders.
+    # holders. This field is specified as a token. For example: "1 token".
```
aterga, line 155: How about `This field is specified as a token amount...`

example_sns_init.yaml
```diff
@@ -133,16 +179,17 @@ Proposals:
...
+    # This field is specified as a duration. For example: "1 day".
     maximum_wait_for_quiet_deadline_extension: 1 day
 
-# Configuration of SNS neurons. These fields have no effect on the NNS proposal,
-# but will affect the SNS itself.
+# Configuration of SNS voting.
 Neurons:
-    # The minimum amount of SNS tokens to stake a neuron.
+    # The minimum amount of SNS tokens to stake a neuron. This field is specified
+    # as a token. For instance, "1 token".
```
aterga, line 189: How about `token amount`?

aterga: Added to NNS1-2508

example_sns_init.yaml
aterga: Added to NNS1-2508

PR #47, opened 2023-09-07
Update example_sns_init.yaml

bjoernek:
Adjustments of the provided example values for the dissolve delay bonus and age bonus.

### DIFF
--- a/example_sns_init.yaml
+++ b/example_sns_init.yaml
@@ -232,10 +232,11 @@ Voting:
             # This field is specified as a duration. For example: "8 years".
             duration: 8 years
             # If a neuron's dissolve delay is `duration`, its voting power will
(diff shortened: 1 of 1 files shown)


DanielThurau:
@bjoernek So I went through the code, and even extended the sns-cli to print out what the equivalent `SnsInitPayload` would be. My findings are as follows:

In the v1 sns init config file, a multiplier of `2.0` would produce a `max_dissolve_delay_bonus_percentage` of 100%. In the v2 sns init config file, a percentage of `100%` would produce a `max_dissolve_delay_bonus_percentage` of 100%. So I believe the example percentage given in the current template is correct. 

Does that make sense to you?

bjoernek:
Based on our discussion the description in the init file is update.

DanielThurau (approved):

PR #55, opened 2023-11-20
Fix `generate_identities_for_neurons_fund_neurons.sh` and improve instructions for Apple Silicon

aterga:
This PR addresses an issue due to which some macOS users were hitting errors while trying to follow README instructions on Apple Silicon systems.

### DIFF
--- a/generate_identities_for_neurons_fund_neurons.sh
+++ b/generate_identities_for_neurons_fund_neurons.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -xeuo pipefail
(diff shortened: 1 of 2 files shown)


generate_identities_for_neurons_fund_neurons.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -xeuo pipefail
```
aterga, line 3:
```suggestion
set -euo pipefail
```

DanielThurau (approved):

PR #57, opened 2023-12-27
Download test wasm instead of recompiling it

aterga:


mraszyk:
Note that this change makes the script unusable with a custom canister name and thus we might as well integrate its functionality into deploy_test_canister.sh

aterga:
> Note that this change makes the script unusable with a custom canister name and thus we might as well integrate its functionality into deploy_test_canister.sh

Thanks for the tip, @mraszyk. I gave it another try.

anchpop (approved):

PR #71, opened 2024-05-24
chore: bump SNS aggregator version and configure it to be fast

mraszyk:
This PR bumps the SNS aggregator version and configures it to be fast. This makes the NNS frontend dapp launchpad update quickly once an NNS proposal to create an SNS is executed.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -78,7 +78,7 @@ then
   curl -L "https://github.com/dfinity/nns-dapp/releases/download/${SNS_AGGREGATOR_RELEASE}/sns_aggregator_dev.wasm.gz" -o nns-dapp/out/sns_aggregator.wasm
 fi
 
-${DFX} canister install sns_aggregator --network "${NETWORK}" --wasm nns-dapp/out/sns_aggregator.wasm '(opt record { update_interval_ms = 1000; fast_interval_ms = 100; })'
+${DFX} canister install sns_aggregator --network "${NETWORK}" --wasm nns-dapp/out/sns_aggregator.wasm '(opt record { update_interval_ms = 1000:nat64; fast_interval_ms = 100:nat64; })'
 ${DFX} canister install internet_identity --network "${NETWORK}" --wasm internet-identity/internet_identity_dev.wasm --argument '(null)'
 ${DFX} canister install nns-dapp --network "${NETWORK}" --wasm nns-dapp/out/nns-dapp.wasm --argument '(opt record{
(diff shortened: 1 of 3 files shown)


setup.sh
```diff
@@ -70,10 +70,15 @@ if [ ! -z "${NNS_DAPP_RELEASE:-}" ]
 then
   mkdir -p nns-dapp/out
   curl -L "https://github.com/dfinity/nns-dapp/releases/download/${NNS_DAPP_RELEASE}/nns-dapp.wasm.gz" -o nns-dapp/out/nns-dapp.wasm
-  curl -L "https://github.com/dfinity/nns-dapp/releases/download/${NNS_DAPP_RELEASE}/sns_aggregator_dev.wasm.gz" -o nns-dapp/out/sns_aggregator.wasm
 fi
 
-${DFX} canister install sns_aggregator --network "${NETWORK}" --wasm nns-dapp/out/sns_aggregator.wasm
+if [ ! -z "${SNS_AGGREGATOR_RELEASE:-}" ]
```
aterga, line 75: Why do we need the if statement?

setup.sh
```diff
@@ -70,10 +70,15 @@ if [ ! -z "${NNS_DAPP_RELEASE:-}" ]
...
+then
+  mkdir -p nns-dapp/out
+  curl -L "https://github.com/dfinity/nns-dapp/releases/download/${SNS_AGGREGATOR_RELEASE}/sns_aggregator_dev.wasm.gz" -o nns-dapp/out/sns_aggregator.wasm
+fi
+
+${DFX} canister install sns_aggregator --network "${NETWORK}" --wasm nns-dapp/out/sns_aggregator.wasm --argument '(opt record { update_interval_ms = 1000:nat64; fast_interval_ms = 100:nat64; })'
```
aterga, line 81: Consider adding a short comment to motivate these values.

setup.sh
mraszyk: There already used to be an if-statement
```
if [ ! -z "${NNS_DAPP_RELEASE:-}" ]
```
covering both the NNS dapp and the SNS aggregator. Since this PR introduces a separate env var for the SNS aggregator, I had to add a new if statement. Now regarding the motivation: the reason is to allow expert people to put their own wasms into nns-dapp/out to install a custom version of the canisters.

setup.sh
mraszyk: Sure, done.

PR #79, opened 2024-09-18
Upgrade DFX to 0.22.0

anchpop:
When upgrading DFX from v0.19.0 to v0.22.0, there was a change in where DFX stores the registry info, so I added a new place for sns-testing to look for it in. I tried it on my mac and on CI and it seems to work.

I left the other places it looked in case any of them are used in some circumstances, but I don't think they are for this version of DFX. (Perhaps they would be useful if someone wanted to use an old version of DFX though.) The dfx start [docs](https://github.com/dfinity/sdk/blob/master/docs/cli-reference/dfx-start.mdx) say that:

> dfx stores data for the shared local network in one of the following locations, depending on your operating system:
> 
> $HOME/.local/share/dfx/network/local (Linux)
> $HOME/Library/Application Support/org.dfinity.dfx/network/local (Macos)

### DIFF
--- a/constants.sh
+++ b/constants.sh
@@ -118,16 +118,6 @@ if [[ "${TESTNET}" == "local" ]]; then
   then
     export REGISTRY_PATH="$(readlink -f "${REGISTRY}")"
   fi
-  REGISTRY="${HOME}/Library/Application Support/org.dfinity.dfx/network/local/state/replicated_state/ic_registry_local_store"
-  if [[ -d "${REGISTRY}" ]]
-  then
-    export REGISTRY_PATH="$(readlink -f "${REGISTRY}")"
-  fi
-  REGISTRY="${HOME}/.local/share/dfx/network/local/state/replicated_state/ic_registry_local_store"
-  if [[ -d "${REGISTRY}" ]]
-  then
-    export REGISTRY_PATH="$(readlink -f "${REGISTRY}")"
-  fi
   # Determine the operating system
   OS_TYPE="$(uname)"
   if [[ "$OS_TYPE" == "Darwin" ]]; then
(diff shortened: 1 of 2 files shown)


constants.sh
```diff
@@ -128,6 +128,25 @@ if [[ "${TESTNET}" == "local" ]]; then
...
+    BASE_PATH="$HOME/Library/Application Support/org.dfinity.dfx/network/local"
+  elif [[ "$OS_TYPE" == "Linux" ]]; then
+    BASE_PATH="$HOME/.local/share/dfx/network/local"
+  else
+    echo "Unsupported OS type: $OS_TYPE"
+    exit 1
+  fi
+
+  # Find the ic_registry_local_store directory within the base path
+  REGISTRY_FOUND=$(find "$BASE_PATH" -type d -name ic_registry_local_store 2>/dev/null | head -n 1)
```
mraszyk, line 143: > `head -n 1`

I would fail if there are more than one match since you might be picking the wrong registry in that case and get unexpected results later.

constants.sh
```diff
@@ -128,6 +128,25 @@ if [[ "${TESTNET}" == "local" ]]; then
   then
     export REGISTRY_PATH="$(readlink -f "${REGISTRY}")"
   fi
```
mraszyk, line 130: I'd remove the original candidate paths as they are certainly not gonna work without `find "$BASE_PATH" -type d -name ic_registry_local_store`.

mraszyk:
> (Perhaps they would be useful if someone wanted to use an old version of DFX though.)

That would only apply to the first candidate path `.dfx/...` since for the other two candidate paths, the `find` command would also succeed for older versions of DFX (although it was not needed back then).

constants.sh
anchpop: Done!

max-dfinity (approved):

constants.sh
```diff
@@ -118,19 +118,33 @@ if [[ "${TESTNET}" == "local" ]]; then
...
-  REGISTRY="${HOME}/.local/share/dfx/network/local/state/replicated_state/ic_registry_local_store"
-  if [[ -d "${REGISTRY}" ]]
-  then
-    export REGISTRY_PATH="$(readlink -f "${REGISTRY}")"
+
+  # Find the ic_registry_local_store directory within the base path
+  REGISTRY_FOUND=$(find "$BASE_PATH" -type d -name ic_registry_local_store 2>/dev/null)
+
+  if [ $(echo "$REGISTRY_FOUND" | wc -l) -gt 1 ]; then
+      echo "Error: Multiple ic_registry_local_store directories found"
```
mraszyk, line 136:
```suggestion
      echo "Error: Multiple ic_registry_local_store directories found; please start your local replica using `dfx start --clean`"
```

anchpop: Sorry, I thought it wouldn't let me merge until I resolved this

PR #91, opened 2025-09-15
Fix comments in sns_init.yaml for clarity

aterga:


sns_tokenomics_analyzer/sns_init.yaml
```diff
@@ -423,15 +423,15 @@ Swap:
...
+    # Must be smaller than or equal to `maximum_direct_participation_icp`.
     minimum_direct_participation_icp: 20_000 tokens
 
     # Maximum amount of ICP from direct participants. If this amount is achieved,
     # the swap will finalize immediately, without waiting for the due date/time;
     # in this case, the swap would be committed if and only if the number of
     # direct participants (`minimum_participants`) is reached (otherwise, it
     # would be aborted).
-    # Must be at least `min_participants * minimum_direct_participation_icp`.
+    # Must be at least `min_participants * minimum_participant_icp`.
```
dylancm4: `min_participants` isn't correct either. It should be:
```
# Must be at least `minimum_participants * minimum_participant_icp`.
```

aterga:
```suggestion
    # Must be at least `minimum_participants * minimum_participant_icp`.
```

dylancm4:
I suggest you also change [this line](https://github.com/dfinity/sns-testing/blob/c80b4c15804012829851f7c88684947c2850d1fa/example_sns_init.yaml#L368):
```
#      2 million    (governance)
```

To this:
```
#      2 million    (treasury)
```

dylancm4:
I suggest you also change [this line](https://github.com/dfinity/sns-testing/blob/main/sns_tokenomics_analyzer/sns_init.yaml#L406C1-L406C37):
```
#     52 million    (governance)
```

To this:
```
#     52 million    (treasury)
```

aterga:
Thanks for the suggestions, @dylancm4 !

Btw, you can add code suggestions directly, which makes it a breeze to accept them. 

You need to find the Github button or use the following 

```
```suggestion

syntax.

dylancm4:
Thanks @aterga. I’d love to use the suggestion feature for lines that weren’t changed in the PR, but the blue "+" only shows up near diff lines, so I added my comments at the PR level instead.

If there’s a way to add a suggestion for `#      2 million    (governance)` in this PR, could you let me know?

aterga:
I think you're right @dylancm4 

I remember this being possible back when we used GitLab

GitHub's suggestion feature seems much weaker, though.

